PR #358, opened 2025-11-11

polsala:
## Rationale

## Why safe to merge

## Test Plan

## Links

## Mock Justification

### DIFF
--- a/utils/nightly-daily-zen-quote-generator/utils/daily-zen-quote-generator/README.md
+++ b/utils/nightly-daily-zen-quote-generator/utils/daily-zen-quote-generator/README.md
@@ -1,0 +1,23 @@
+# Daily Zen Quote Generator
+
+A whimsical yet useful utility that prints a *Zen* quote of the day.
+
+## Features
+- No external dependencies or network access.
+- Deterministic: the same date always yields the same quote.
+- Simple CLI: `python -m daily_zen_quote_generator` prints the quote.
+- Fully tested with offline, deterministic mocks.
+
+## Usage
+```bash
+cd utils/daily-zen-quote-generator
+python -m src.main
+```
+
+## How it works
+The utility contains a small list of Zen‑style quotes. The quote for a given day is selected by hashing the ISO‑format date (`YYYY‑MM‑DD`) and using the result to index the list.
+
+## Testing
+```bash
+python -m unittest discover -s tests
+```
--- a/utils/nightly-daily-zen-quote-generator/utils/daily-zen-quote-generator/src/main.py
+++ b/utils/nightly-daily-zen-quote-generator/utils/daily-zen-quote-generator/src/main.py
@@ -1,0 +1,52 @@
+import hashlib
+import datetime
+from typing import List
+
+# Embedded list of Zen quotes (feel free to extend)
+_QUOTES: List[str] = [
+    "The journey of a thousand miles begins with one step.",
+    "When the mind is still, the universe surrenders.",
+    "Silence is a source of great strength.",
+    "The obstacle is the path.",
+    "Let go or be dragged.",
+    "In the middle of difficulty lies opportunity.",
+    "A single moment can change a lifetime.",
+    "Nature does not hurry, yet everything is accomplished.",
+    "The only constant is change.",
+    "Know the rules so you can break them wisely."
+]
+
+
+def _hash_date(date_str: str) -> int:
+    """Return a stable integer hash for a given ISO date string.
+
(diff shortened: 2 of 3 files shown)


polsala:
## ✅ What’s solid
- **Self‑contained implementation** – The utility only relies on the Python standard library (`hashlib`, `datetime`, `typing`). No third‑party dependencies or network calls are introduced, which keeps the attack surface minimal.  
- **Deterministic quote selection** – Using a SHA‑256 hash of the ISO‑formatted date guarantees that the same date always maps to the same quote, which is exactly what the README promises.  
- **Clear CLI entry point** – The `if __name__ == "__main__": main()` guard makes the module runnable with `python -m src.main`, matching the usage example.  
- **Simple, readable code** – Functions are short, type‑annotated, and have docstrings that explain the intent and parameters.  

## 🧪 Tests
- **Positive coverage** – The two tests exercise the primary public API (`get_quote_of_day`) both with a deterministic date and with the default “today” path.  
- **Actionable improvements**
  1. **Patch the correct import target**  
     ```python
     # Current (fragile) approach
     with patch.object(datetime.date, "today", return_value=fixed_date):
         result = get_quote_of_day()
     ```
     The `datetime.date.today` call inside `src.main` is resolved at import time, so patching the class directly can be flaky. Prefer patching the attribute *as used in the module*:
     ```python
     with patch("src.main.datetime.date.today", return_value=fixed_date):
         result = get_quote_of_day()
     ```
  2. **Avoid re‑implementing private logic**  
     The test recomputes the expected index by duplicating the hash algorithm. This couples the test to the implementation and defeats the purpose of a black‑box test. Either:
     - Export `_hash_date` (e.g. rename to `hash_date` and add it to `__all__`), or
     - Use a known‑good fixture value:
       ```python
       EXPECTED_QUOTE = "Silence is a source of great strength."
       self.assertEqual(get_quote_of_day(fixed_date), EXPECTED_QUOTE)
       ```
  3. **Test edge cases**  
     - Verify behavior when the quote list is empty (should raise a clear exception).  
     - Confirm that the function works with a non‑UTC date object (e.g., a `datetime.datetime` with a timezone) if you ever extend the API.  
  4. **Test the CLI** – Add a small integration test that spawns the module with `subprocess` and asserts that the output ends with a newline and is one of the known quotes.

## 🔒 Security
- **No secrets or external I/O** – The utility does not read environment variables, files, or network resources, so there is nothing sensitive to leak.  
- **Input validation** – The only external input is the optional `today` argument. Since the function expects a `datetime.date` (or `None`), type hints already guide correct usage. Consider adding a defensive check:
  ```python
  if today is not None and not isinstance(today, datetime.date):
      raise TypeError("today must be a datetime.date instance")
  ```
- **Deterministic hashing** – Using SHA‑256 on a controlled string (`YYYY‑MM‑DD`) is safe; there is no risk of hash‑collision attacks affecting functionality.

- **README path mismatch** – The usage snippet shows:
  ```bash
  cd utils/daily-zen-quote-generator
  python -m src.main
  ```
  This works only when the current working directory is the *package* root. A more conventional entry point would be:
  ```bash
  cd utils/daily-zen-quote-generator
  python -m daily_zen_quote_generator
  ```
  To enable that, add an `__init__.py` in `src/` and expose `main` via a `__main__.py`, or adjust the README to reflect the exact import path.
- **Packaging hints** – If this utility is intended to be reusable across the monorepo, consider adding a minimal `pyproject.toml` or `setup.cfg` so it can be installed with `pip -e .`. This also makes the CLI discoverable via an entry point (`[project.scripts] daily-zen = "daily_zen_quote_generator.src.main:main"`).  
- **Versioning** – The generated folder lives under `utils/nightly-daily-zen-quote-generator`. If the utility evolves, a simple `__version__ = "0.1.0"` constant and a changelog entry would help downstream users track updates.

## 🧱 Mocks / Fakes
- **Current mock strategy** – Patching `datetime.date.today` is appropriate for deterministic unit tests and keeps them offline.  
- **Alternative: freezegun** – For readability and broader date‑time control, the `freezegun` library can freeze the entire `datetime` module:
  ```python
  from freezegun import freeze_time

  @freeze_time("2023-10-31")
  def test_known_date_returns_expected_quote(self):
      self.assertEqual(get_quote_of_day(), EXPECTED_QUOTE)
  ```
  This eliminates the need for `patch.object` and makes the intent explicit.  
- **Future mock needs** – If the utility ever expands to fetch external quotes or read configuration files, introduce a thin abstraction layer (e.g., a `QuoteProvider` protocol) that can be swapped with a fake implementation in tests. This will keep the test suite fast and deterministic.

PR #598, opened 2025-11-14

polsala:
## Rationale

## Why safe to merge

## Test Plan

## Links

## Mock Justification

### DIFF
--- a/utils/nightly-human-duration-parser/utils/human-duration-parser/README.md
+++ b/utils/nightly-human-duration-parser/utils/human-duration-parser/README.md
@@ -1,0 +1,47 @@
+# Human Duration Parser
+
+A tiny, zero‑dependency utility that converts human‑readable duration strings into a total number of seconds.
+
+## Features
+
+- Supports days (`d`), hours (`h`), minutes (`m`), and seconds (`s`).
+- Order‑agnostic (e.g., `5h2d` works the same as `2d5h`).
+- Simple Python API: `parse_duration("1d2h30m") -> 93900`.
+- Small command‑line interface `duration-to-seconds`.
+
+## Installation
+
+Just copy the folder `utils/human-duration-parser/` into your project and import:
+
+```python
+from utils.human-duration-parser.src.parser import parse_duration
+```
+
+## Usage
+
+### Library
+
+```python
+from utils.human-duration-parser.src.parser import parse_duration
+
+seconds = parse_duration("1d 2h 30m")
+print(seconds)  # 93900
+```
+
+### CLI
+
+```bash
+$ python -m utils.human-duration-parser.src.parser "3h45m"
+13500
+```
+
+## Supported format
+
+- Numbers must be integers (no floats).
+- Units: `d` (days), `h` (hours), `m` (minutes), `s` (seconds).
+- Whitespace is ignored.
+- Example valid strings: `"2d"`, `"5h30m"`, `"45m10s"`, `"1d 2h 3m 4s"`.
+
+## Error handling
+
+Invalid strings raise a `ValueError` with a helpful message.
--- a/utils/nightly-human-duration-parser/utils/human-duration-parser/src/parser.py
+++ b/utils/nightly-human-duration-parser/utils/human-duration-parser/src/parser.py
@@ -1,0 +1,74 @@
+import argparse
+import re
+import sys
+from typing import Dict
+
+# Regex pattern to capture optional groups for days, hours, minutes, seconds
+_DURATION_RE = re.compile(
+    r"(?:(?P<days>\d+)\s*d)?\s*"
+    r"(?:(?P<hours>\d+)\s*h)?\s*"
+    r"(?:(?P<minutes>\d+)\s*m)?\s*"
+    r"(?:(?P<seconds>\d+)\s*s)?\s*$",
+    re.IGNORECASE,
+)
+
+_UNIT_TO_SECONDS: Dict[str, int] = {
+    "days": 86400,
+    "hours": 3600,
+    "minutes": 60,
+    "seconds": 1,
+}
+
+
+def parse_duration(duration_str: str) -> int:
+    """Parse a human‑friendly duration string into total seconds.
+
+    Supported units are days (d), hours (h), minutes (m), and seconds (s).
+    The order of units does not matter and whitespace is ignored.
+
+    Args:
+        duration_str: String like "1d 2h30m" or "45m10s".
+
+    Returns:
+        Total number of seconds as an integer.
+
+    Raises:
+        ValueError: If the string cannot be parsed.
+    """
+    if not isinstance(duration_str, str):
+        raise ValueError("duration_str must be a string")
+
+    # Remove all whitespace for easier matching
+    cleaned = duration_str.replace(" ", "")
+    match = _DURATION_RE.fullmatch(cleaned)
+    if not match:
+        raise ValueError(f"Unable to parse duration string: '{duration_str}'")
+
+    total_seconds = 0
+    for unit, factor in _UNIT_TO_SECONDS.items():
+        value = match.group(unit)
+        if value is not None:
+            total_seconds += int(value) * factor
+    return total_seconds
+
+
+def _cli() -> None:
+    parser = argparse.ArgumentParser(
+        description="Convert a human‑friendly duration string to seconds."
+    )
+    parser.add_argument(
+        "duration",
+        type=str,
+        help="Duration string (e.g., '2d5h30m').",
+    )
+    args = parser.parse_args()
+    try:
+        seconds = parse_duration(args.duration)
+        print(seconds)
+    except ValueError as exc:
+        print(f"Error: {exc}", file=sys.stderr)
+        sys.exit(1)
+
+
+if __name__ == "__main__":
+    _cli()
--- a/utils/nightly-human-duration-parser/utils/human-duration-parser/tests/test_parser.py
+++ b/utils/nightly-human-duration-parser/utils/human-duration-parser/tests/test_parser.py
@@ -1,0 +1,32 @@
+import unittest
+from utils.human-duration-parser.src.parser import parse_duration
+
+class TestParseDuration(unittest.TestCase):
+    def test_simple_cases(self):
+        self.assertEqual(parse_duration("1d"), 86400)
+        self.assertEqual(parse_duration("2h"), 7200)
+        self.assertEqual(parse_duration("30m"), 1800)
+        self.assertEqual(parse_duration("45s"), 45)
+
+    def test_combined_cases(self):
+        self.assertEqual(parse_duration("1d2h30m"), 86400 + 7200 + 1800)
+        self.assertEqual(parse_duration("2h45m10s"), 7200 + 2700 + 10)
+        self.assertEqual(parse_duration("5m 3d"), 5 * 60 + 3 * 86400)
+        self.assertEqual(parse_duration("  4h   5s  "), 4 * 3600 + 5)
+
+    def test_invalid_inputs(self):
+        with self.assertRaises(ValueError):
+            parse_duration("")
+        with self.assertRaises(ValueError):
+            parse_duration("abc")
+        with self.assertRaises(ValueError):
+            parse_duration("10x")  # unsupported unit
+        with self.assertRaises(ValueError):
+            parse_duration("1.5h")  # floats not allowed
+
+    def test_non_string_input(self):
+        with self.assertRaises(ValueError):
+            parse_duration(123)  # type: ignore
+
+if __name__ == "__main__":
+    unittest.main()


polsala:
## ✅ What’s solid
- **Clear separation of concerns** – the parser, CLI wrapper, and documentation live in their own folder, making the utility easy to drop into other projects.  
- **Zero‑dependency implementation** – only the standard library is used, which keeps the footprint tiny and avoids supply‑chain risk.  
- **Helpful error messages** – `ValueError` includes the offending string, which is great for debugging.  
- **Comprehensive README** – the feature list, installation steps, and usage examples (both library and CLI) are well‑written and immediately usable.  
- **Test suite** – basic happy‑path and failure cases are covered, and the test file follows the repository’s existing `unittest` style.

---

## 🧪 Tests
- **Good baseline coverage** – the current tests hit simple units, combined strings, whitespace handling, and invalid inputs.  
- **Add order‑agnostic verification** – the README claims the parser works regardless of unit order, but the regex currently enforces `days → hours → minutes → seconds`. Add a test such as:

  ```python
  def test_order_agnostic(self):
      self.assertEqual(parse_duration("5h2d"), 2*86400 + 5*3600)
  ```

- **Duplicate unit detection** – currently `"1h2h"` would silently ignore the second occurrence. A test that expects a `ValueError` will surface this edge case:

  ```python
  with self.assertRaises(ValueError):
      parse_duration("1h2h")
  ```

- **Boundary values** – consider testing very large numbers (e.g., `"9999d"`) and leading zeros (`"01h"`). This ensures integer conversion and overflow handling behave as expected.  
- **Parametrized testing** – switching to `pytest` with `@pytest.mark.parametrize` would make the test matrix more concise and easier to extend.

---

## 🔒 Security
- **No external I/O** – the parser only works on the supplied string, so there’s no surface for injection attacks.  
- **CLI error handling** – errors are printed to `stderr` and the process exits with a non‑zero status, which is appropriate for scripting contexts.  
- **Future hardening tip** – if the utility ever expands to accept input from files or environment variables, validate length and character set early to avoid denial‑of‑service via extremely long strings.

---

## 🧩 Docs/DX
- **Installation note** – “Just copy the folder” works, but many users expect a pip‑installable package. Adding a minimal `setup.py`/`pyproject.toml` (or at least a note on how to add the folder to `PYTHONPATH`) would improve discoverability.  
- **Import path clarification** – the README shows `from utils.human-duration-parser.src.parser import parse_duration`. Because hyphens are not valid in Python package names, this works only when the folder is on the import path. Suggest renaming the package directory to `human_duration_parser` (or using an underscore) and updating the import accordingly.  
- **Correct the “order‑agnostic” claim** – either adjust the implementation (see below) or update the docs to reflect the current limitation.  
- **CLI usage example** – include the `-h/--help` output in the README to demonstrate available flags.  

---

## 🧱 Mocks/Fakes
- **Not required for current scope** – the parser is pure and deterministic, so no mocks are needed.  
- **Potential future need** – if you later add time‑based features (e.g., “parse relative to now”), consider introducing a small wrapper around `datetime.now()` that can be mocked in tests.

---

### Quick win: make the parser truly order‑agnostic
Replace the single‑pass regex with a pattern that finds *any* unit/value pair, then aggregate:

```python
_UNIT_RE = re.compile(r"(?P<value>\d+)\s*(?P<unit>[dhms])", re.IGNORECASE)

def parse_duration(duration_str: str) -> int:
    if not isinstance(duration_str, str):
        raise ValueError("duration_str must be a string")

    cleaned = duration_str.replace(" ", "")
    if not cleaned:
        raise ValueError("duration string is empty")

    total = 0
    seen = set()
    for m in _UNIT_RE.finditer(cleaned):
        unit = m.group("unit").lower()
        if unit in seen:
            raise ValueError(f"Duplicate unit '{unit}' in '{duration_str}'")
        seen.add(unit)
        value = int(m.group("value"))
        total += value * _UNIT_TO_SECONDS[unit + ("s" if unit != "s" else "")]
    # Ensure the whole string was consumed
    if _UNIT_RE.sub("", cleaned):
        raise ValueError(f"Unrecognised parts in '{duration_str}'")
    return total
```

This implementation:

- Accepts units in any order.  
- Detects duplicate units and raises a clear error.  
- Fails fast on stray characters (e.g., `"10hxyz"`).  

Add corresponding tests (order‑agnostic, duplicate detection, stray characters) and the utility will fully match its advertised behavior.

polsala:
## 🤖 Review by GEMINI Agent

✅ What's solid
*   The utility demonstrates excellent separation of concerns, with the parser logic, CLI wrapper, and documentation residing in a self-contained folder. This design promotes reusability and simplifies integration into other projects.
*   The implementation is zero-dependency, relying solely on the Python standard library. This minimizes supply-chain risk and keeps the utility's footprint exceptionally small.
*   Error handling is robust, with `ValueError` exceptions providing clear, actionable messages that include the offending input string, aiding in debugging.
*   The regular expression (`_DURATION_RE`) is well-crafted, effectively handling various valid duration formats, including order-agnostic units and flexible whitespace, while `re.IGNORECASE` adds user-friendliness.
*   The CLI provides a practical command-line interface, and its error output correctly uses `sys.stderr` for diagnostic messages.

🧪 Tests
*   The test suite covers a good range of scenarios, including simple unit conversions, combined durations, and various invalid inputs (empty, non-numeric, unsupported units, floats).
*   Consider adding test cases for:
    *   Zero-value durations, e.g., `"0d"`, `"0h0m0s"`, or an empty string after cleaning (which currently raises `ValueError`).
    *   Strings containing only whitespace, e.g., `"   "`, to ensure consistent `ValueError` handling.
    *   Mixed-case unit inputs, such as `"1D 2h 3M 4s"`, to explicitly verify the `re.IGNORECASE` flag's behavior.
    *   Very large integer inputs for units (e.g., `1000000000d`) to confirm Python's arbitrary-precision integers are handled as expected, even if not explicitly constrained by the problem domain.

🔒 Security
*   The utility's self-contained nature and lack of external dependencies significantly reduce its attack surface and eliminate common supply-chain vulnerabilities.
*   The regular expression used (`_DURATION_RE`) is straightforward and does not exhibit patterns prone to Regular Expression Denial of Service (ReDoS) attacks, such as excessive backtracking.
*   Input validation is appropriately handled through type checking (`isinstance`) and strict regex matching (`fullmatch`), preventing unexpected data types or malformed strings from causing runtime issues.
*   The CLI leverages `argparse` for argument parsing, which is a secure and standard way to handle command-line inputs, mitigating injection risks.

🧩 Docs/DX
*   The `README.md` is comprehensive, clearly outlining features, installation, usage examples (both library and CLI), supported formats, and error handling.
*   The `parse_duration` function includes a clear docstring with detailed explanations of its purpose, arguments, return value, and potential exceptions.
*   The installation instructions in the `README.md` suggest copying the folder. However, the example import path `from utils.human-duration-parser.src.parser import parse_duration` uses hyphens, which are invalid in Python module names. This should be corrected to use underscores (e.g., `human_duration_parser`) in the example, assuming the user renames the folder/module accordingly after copying.
*   Consider adding a docstring to the private `_cli` function to explain its role as the command-line entry point.
*   Brief comments for the `_DURATION_RE` and `_UNIT_TO_SECONDS` global variables could enhance clarity for future maintainers.

🧱 Mocks/Fakes
*   Mocks or fakes are not applicable to this utility. The `parse_duration` function is a pure function, operating solely on its input arguments without external dependencies, side effects, or complex state.
*   The CLI wrapper is minimal and directly calls the pure parsing function, eliminating the need for mocking external interactions. This design choice contributes to the utility's simplicity and testability.

polsala:
## 🤖 Review by GROQ Agent

<think>
Okay, let's start by looking at the PR. The title is "feat(openrouter): autogenerated update", and the rationale mentions an automated proposal from the OpenRouter generator. The main files added are a README, a parser.py, and test_parser.py. 

First, the ✅ What's solid section. The user mentioned separation of concerns and zero dependencies. The code is in a self-contained folder, which is good. The parser uses only the standard library, which reduces external dependencies. Also, the error messages include the offending string, which helps with debugging. Those are solid points.

Next, the 🧪 Tests. The test cases cover simple and combined durations, whitespace handling, invalid inputs, and non-string inputs. That's comprehensive. But wait, the test for "1.5h" is there, but the code currently raises an error for floats. However, the regex might not catch all cases. For example, if someone inputs "1d 2h 3m 4s" with varying whitespace, the test covers that. But maybe edge cases like leading or trailing whitespace aren't tested. Also, the test for non-string input is there, but maybe more edge cases could be added, like other types like None or a list.

🔒 Security: The code uses regex with re.IGNORECASE, which is okay. But the regex pattern might have some issues. Let me check the regex. The pattern is r"(?:(?P<days>\d+)\s*d)?\s*...". Wait, the regex uses optional groups for each unit, but the order is fixed as days, hours, minutes, seconds. The README says it's order-agnostic, but the regex is written in a specific order. Wait, no—the regex allows for optional groups in that order, but the user can input any combination. Wait, no, the regex is written to match days first, then hours, etc., but each is optional. So if someone writes "2h5d", the regex would still capture days and hours because they're optional. Wait, no—the regex is structured as (?:...)? for each unit. So the order in the regex is days, then hours, then minutes, then seconds. But the user can input any order as long as the units are present. For example, "5h30m" would match because the days and minutes parts are optional. Wait, but the regex is written to look for days first, then hours, etc. So if the input is "30m5h", the regex would capture hours and minutes correctly? Let me think. The regex for hours is (?:\d+\s*h)?\s*, so if the input is "30m5h", the regex would first look for days (none), then hours (none), then minutes (30), then seconds (none). Wait, but the input has "30m5h"—the regex is written to capture days first, then hours, then minutes, then seconds. So in this case, the regex would not capture the '5h' because after the minutes (30m), the hours part is already passed. Wait, that's a problem. The regex is structured in a fixed order, so if the input has hours after minutes, the regex would miss it. For example, "5h30m" would work because the hours are first, but "30m5h" would not match because the regex expects hours to come before minutes. Oh no, that's a big issue. The README says it's order-agnostic, but the regex is not. So this is a security or correctness issue. The code claims to handle order-agnostic input, but the regex is not written to do that. That's a problem. So the security section should mention that the regex is not order-agnostic as claimed, leading to potential incorrect parsing.

🧩 Docs/DX: The README is good, but the installation instruction says to copy the folder into the project. However, the import path in the example is "utils.human-duration-parser.src.parser", but the actual file structure is "utils/nightly-human-duration-parser/utils/human-duration-parser/src/parser.py". So if someone copies the folder into their project, they might have to adjust the import path. Also, the CLI example uses "python -m utils.human-duration-parser.src.parser", but the actual module structure might require a different way to run it. The documentation might be misleading here. Also, the README mentions the CLI but doesn't explain how to install it as a package or how to use it outside of the project structure. So the documentation could be improved for better developer experience.

🧱 Mocks/Fakes: The PR mentions that no mocks were introduced, which is correct. The tests are unit tests without mocks. However, the test for the CLI is not present. The _cli() function is tested indirectly through the unittest, but there's no test that actually runs the CLI and checks the output. The current tests only test the parse_duration function. So the CLI functionality is not tested. That's a gap in testing.

Putting it all together, the review should highlight the regex issue in the security section, the missing CLI tests in the tests section, the documentation issues in the docs/DX section, and the lack of mocks where needed (though none were added here). Also, the tests could be improved with more edge cases.
</think>

```markdown
## ✅ What's solid
- **Modular structure** – The utility is self-contained in a dedicated folder, promoting reusability and isolation from the main codebase.
- **Zero external dependencies** – Relies solely on Python's standard library (`re`, `argparse`, etc.), minimizing supply-chain risks and simplifying deployment.
- **Clear error handling** – Raises `ValueError` with the original input string for invalid cases, aiding debugging (e.g., `ValueError: Unable to parse duration string: '10x'`).

---

## 🧪 Tests
- **Comprehensive unit tests** – Covers valid/invalid inputs, whitespace handling, and type checks (e.g., `test_invalid_inputs` for unsupported units like `x`).
- **Missing edge cases**:
  - Test for **leading/trailing whitespace** in inputs (e.g., `"  1d 2h 3m 4s  "`).
  - Test for **partial matches** (e.g., `"1d2h"` without minutes/seconds).
  - Test for **maximum integer values** to avoid overflow (though unlikely in Python).
- **CLI testing gap** – The `_cli()` function is untested directly. Consider adding a test that invokes the CLI via `subprocess` to validate exit codes and output.

---

## 🔒 Security
- **Regex vulnerability** – The current regex pattern (`_DURATION_RE`) assumes a fixed order of units (`d → h → m → s`). This contradicts the README's claim of "order-agnostic" parsing. For example:
  ```python
  parse_duration("5h2d")  # Fails to match due to regex order
  ```
  **Fix**: Use a regex that captures all units regardless of order. Example:
  ```python
  _DURATION_RE = re.compile(
      r"(?:(?P<days>\d+)\s*d)?\s*"
      r"(?:(?P<hours>\d+)\s*h)?\s*"
      r"(?:(?P<minutes>\d+)\s*m)?\s*"
      r"(?:(?P<seconds>\d+)\s*s)?\s*$",
      re.IGNORECASE,
  )
  ```
  This allows any order but still requires units to appear in the regex sequence. A more robust solution would use a regex that captures all units in any order (e.g., using named groups with `|` separators).

- **Input sanitization** – The regex does not reject inputs with **non-integer numbers** (e.g., `"1.5h"`). While the test `test_invalid_inputs` catches this, the regex could explicitly disallow decimal points:
  ```python
  r"(?P<days>\d+)"  # Instead of \d* to enforce at least one digit
  ```

---

## 🧩 Docs/DX
- **Installation instructions** – The README suggests importing via `from utils.human-duration-parser.src.parser import parse_duration`, but the actual file path is `utils/nightly-human-duration-parser/utils/human-duration-parser/src/parser.py`. This may confuse users who copy the folder directly.
- **CLI usage** – The CLI example (`python -m utils.human-duration-parser.src.parser "3h45m"`) assumes a specific project structure. Consider adding a note about installing the utility as a package or adjusting `PYTHONPATH` for easier testing.
- **Documentation clarity** – The "Supported format" section should explicitly mention that **units must be in the order `d`, `h`, `m`, `s`** due to the current regex implementation, to avoid misleading users about "order-agnostic" behavior.

---

## 🧱 Mocks/Fakes
- **No mocks required** – The utility is stateless and does not interact with external systems, so mocks are unnecessary.
- **CLI testing recommendation** – Add a test using `subprocess.run()` to validate the CLI interface:
  ```python
  def test_cli_output():
      result = subprocess.run(
          [sys.executable, "-m", "parser", "1d2h30m"],
          capture_output=True,
          text=True,
      )
      assert result.returncode == 0
      assert result.stdout.strip() == "93900"
  ```
  This ensures the CLI behaves as documented.
```

PR #659, opened 2025-11-15

polsala:
## Rationale

## Why safe to merge

## Test Plan

## Links

## Mock Justification

### DIFF
--- a/utils/nightly-daily-haiku-generator/utils/daily-haiku-generator/README.md
+++ b/utils/nightly-daily-haiku-generator/utils/daily-haiku-generator/README.md
@@ -1,0 +1,27 @@
+# Daily Haiku Generator
+
+A tiny, self‑contained utility that produces a three‑line haiku (5‑7‑5 syllable pattern) based on the current date or any supplied date.
+
+## Features
+- Deterministic output: the same date always yields the same haiku.
+- No external dependencies – pure Python 3.11.
+- Simple CLI (`python -m src.haiku`) prints today’s haiku.
+- Library function `generate(date: datetime.date | None) -> str` for programmatic use.
+
+## Usage
+```bash
+# Print today's haiku
+python -m src.haiku
+```
+```python
+from src.haiku import generate
+import datetime
+
+print(generate(datetime.date(2023, 1, 1)))
+```
+
+## Testing
+Run the bundled tests with:
+```bash
+python -m unittest discover -s tests
+```
--- a/utils/nightly-daily-haiku-generator/utils/daily-haiku-generator/src/haiku.py
+++ b/utils/nightly-daily-haiku-generator/utils/daily-haiku-generator/src/haiku.py
@@ -1,0 +1,49 @@
+'''Daily Haiku Generator utility.
+
+Provides a function to generate a whimsical haiku based on a date.
+''' 
+
+import random
+import datetime
+from typing import Optional
+
+# Word banks for haiku lines (5‑7‑5 syllable patterns)
+FIVE_SYLLABLE_LINES = [
+    "Morning dew glistens",
+    "Silent moonlit night",
+    "Leaves whisper softly",
+    "Snowflakes kiss the earth",
+    "Stars flicker above",
+]
+
+SEVEN_SYLLABLE_LINES = [
+    "A breeze carries distant laughter",
+    "Dreams drift on the river of time",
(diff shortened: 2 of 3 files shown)


polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **Clear isolation** – The whole utility lives under its own `utils/nightly-daily-haiku-generator` folder, so it won’t interfere with other parts of the repo.  
- **Deterministic output** – Seeding `random.Random` with `date.toordinal()` guarantees the same date always yields the same haiku. This makes the function pure and easy to test.  
- **Zero external dependencies** – Pure‑Python 3.11 code, no third‑party packages, which keeps the attack surface tiny and the install footprint negligible.  
- **Simple CLI entry‑point** – `python -m src.haiku` works out‑of‑the‑box for quick manual checks.  
- **Good docstring** – The module‑level docstring explains the purpose and the public API (`generate`).  

## 🧪 Tests
- **Coverage of core contract** – The three tests verify the most important guarantees:
  1. Exactly three lines are returned.  
  2. The same date always produces the same haiku.  
  3. Each line comes from the predefined word banks.  

- **Actionable improvements**
  - **Test the default (today) path** – Currently `generate()` without an argument isn’t exercised. Patch `datetime.date.today` with `unittest.mock` to assert that the function uses the current date.
    ```python
    @patch('src.haiku.datetime.date')
    def test_default_uses_today(self, mock_date):
        mock_date.today.return_value = datetime.date(2024, 1, 1)
        haiku = generate()
        # assert haiku matches the seed for 2024‑01‑01
    ```
  - **Parameter validation** – Add a test that passing a non‑date (e.g., a string) raises a clear `TypeError`. This encourages defensive programming.
  - **Edge‑case dates** – Verify behavior for leap‑day (`2020‑02‑29`) and far‑future dates to ensure `toordinal()` never overflows.
  - **Word‑bank sanity** – A quick sanity test that each bank contains at least one line and that every line actually follows the 5‑/7‑syllable pattern (optional but helpful for future contributors).
    ```python
    def test_wordbanks_nonempty(self):
        self.assertTrue(FIVE_SYLLABLE_LINES)
        self.assertTrue(SEVEN_SYLLABLE_LINES)
    ```

## 🔒 Security
- **No external input handling** – The only user‑controlled value is the optional `date` argument, which is a `datetime.date` object. There’s no string interpolation or filesystem access, so injection risks are essentially nil.  
- **Deterministic randomness** – Because the RNG is seeded locally and never exposed, there’s no chance of leaking entropy or relying on a global random state.  
- **CLI safety** – The script only prints the generated haiku; it does not read from stdin or write to disk, so the command line surface is safe.  
- **Potential hardening**  
  - Add a type check (or `isinstance(date, datetime.date)`) and raise a clear exception if the caller supplies an incompatible type. This prevents accidental misuse that could otherwise surface as obscure errors.  

- **README completeness** – The README covers features, usage, and testing, which is great for newcomers.  
- **Minor enhancements**
  - **Installation tip** – Mention that the utility can be run directly from the repo root or installed as a module (e.g., `pip install -e .` if you expose a package).  
  - **CLI invocation clarification** – The command `python -m src.haiku` works only when the current working directory is the `src` folder. Suggest a more robust entry point, such as adding a console script entry in `setup.cfg` or using `python -m utils.daily_haiku_generator.src.haiku`.  
  - **Example with custom date string** – Show how to parse a string into a `datetime.date` before calling `generate`, which helps users unfamiliar with the `datetime` API.  
  - **Syllable‑count disclaimer** – The word banks are curated to follow the 5‑7‑5 pattern, but the code does not enforce it. A short note in the README can set expectations and invite contributors to add a validator if they wish.  

## 🧱 Mocks / Fakes
- **Current state** – No mocks are required for the deterministic logic, which keeps the test suite lightweight.  
- **Suggested mock usage**
  - **Patch `datetime.date.today`** (as shown above) to test the default‑date path without relying on the actual system clock.  
  - **Future extensibility** – If the utility ever expands to fetch external word banks (e.g., from a file or API), introduce a thin abstraction (e.g., `WordBankProvider`) that can be faked in tests. This will keep the current deterministic tests valid while allowing richer functionality later.  

---  

Overall, the utility is well‑scoped, safe, and comes with a solid baseline test suite. The suggestions above are optional refinements that will make the codebase more robust, easier to extend, and clearer for future contributors.

PR #836, opened 2025-11-18

polsala:
## Rationale

## Why safe to merge

## Test Plan

## Links

## Mock Justification

### DIFF
--- a/utils/nightly-nightly-emoji-forecast/utils/nightly-emoji-forecast/README.md
+++ b/utils/nightly-nightly-emoji-forecast/utils/nightly-emoji-forecast/README.md
@@ -1,0 +1,22 @@
+# Nightly Emoji Forecast
+
+`nightly-emoji-forecast` is a tiny, self‑contained Python utility that returns a single emoji representing the current weather for a given city.
+
+## Features
+- Maps common weather conditions to clear, expressive emojis.
+- Offline‑friendly: the core logic is pure Python; the external weather fetch is deliberately left as a stub so it can be mocked in tests.
+- Simple CLI: `python -m utils.nightly_emoji_forecast.src.forecast <city>` prints `<city>: <emoji>`.
+
+## Installation & Usage
+```bash
+# From the repository root
+python -m utils.nightly_emoji_forecast.src.forecast "San Francisco"
+```
+
+If you integrate a real weather API, replace the `_fetch_weather` stub in `forecast.py` with your implementation.
+
+## Testing
+```bash
+python -m unittest discover utils/nightly-emoji-forecast/tests
+```
+All tests run offline using `unittest.mock`.
--- a/utils/nightly-nightly-emoji-forecast/utils/nightly-emoji-forecast/src/forecast.py
+++ b/utils/nightly-nightly-emoji-forecast/utils/nightly-emoji-forecast/src/forecast.py
@@ -1,0 +1,57 @@
+"""
+emoji_forecast utility
+
+Provides a quick emoji representation of the current weather for a city.
+"""
+
+import argparse
+from typing import Dict
+
+# Mapping of weather condition keywords to emojis
+_WEATHER_EMOJI_MAP: Dict[str, str] = {
+    "clear": "☀️",
+    "sunny": "☀️",
+    "cloudy": "☁️",
+    "overcast": "☁️",
+    "rain": "🌧️",
+    "drizzle": "🌦️",
+    "thunderstorm": "⛈️",
+    "snow": "❄️",
+    "mist": "🌫️",
+    "fog": "🌫️",
+    "unknown": "❓",
+}
+
+
+def _fetch_weather(city: str) -> str:
+    """Placeholder for real weather fetching logic.
+
+    Returns a lowercase weather condition string.
+    In production this could call an API like OpenWeatherMap.
+    """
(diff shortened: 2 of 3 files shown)


polsala:
## ✅ What’s solid
- **Self‑contained utility** – All code lives under a single `utils/...` folder, keeping the rest of the repo untouched.  
- **Clear separation of concerns** – The weather‑fetching stub (`_fetch_weather`) is isolated from the emoji‑mapping logic, making it easy to replace with a real API later.  
- **Typed mapping** – `_WEATHER_EMOJI_MAP` is declared with `Dict[str, str]`, which helps static analysis tools.  
- **CLI entry point** – The `if __name__ == "__main__": main()` block provides a quick way to run the tool locally.  
- **README** – Gives a concise description, feature list, installation snippet, and test command.

## 🧪 Tests
- **Dynamic module loading** works for a non‑package layout, but the current patching strategy is broken.  
  ```python
  @patch.object(load_forecast_module, "_fetch_weather")
  ```
  `load_forecast_module` is a *function*, not the module that contains `_fetch_weather`. This raises an `AttributeError` and the tests will never hit the real logic.

- **Suggested fix** – Load the module once in `setUp` and patch the attribute on that instance:
  ```python
  class TestEmojiForecast(unittest.TestCase):
      def setUp(self):
          self.forecast = load_forecast_module()

      @patch.object(self.forecast, "_fetch_weather")
      def test_clear_weather(self, mock_fetch):
          mock_fetch.return_value = "Clear"
          self.assertEqual(self.forecast.get_emoji_forecast("Atlantis"), "☀️")
  ```

- **Package‑style import** – Consider turning the utility into a proper package (add `__init__.py` files) so tests can simply `from utils.nightly_emoji_forecast.src import forecast` and patch with `patch('utils.nightly_emoji_forecast.src.forecast._fetch_weather')`. This removes the need for the custom loader and makes the test suite more idiomatic.

- **Coverage** – Current tests cover the happy path for “clear” and “rain” and the fallback on exception. Adding a parametric test for all keys in `_WEATHER_EMOJI_MAP` would guarantee the mapping stays in sync:
  ```python
  @parameterized.expand([
      ("clear", "☀️"),
      ("cloudy", "☁️"),
      # … all other pairs …
  ])
  def test_mapping(self, condition, expected):
      with patch.object(self.forecast, "_fetch_weather", return_value=condition):
          self.assertEqual(self.forecast.get_emoji_forecast("City"), expected)
  ```

## 🔒 Security
- **No external secrets** – The stub does not contact any API, so there are no credential leaks.  
- **Exception handling** – `get_emoji_forecast` catches *any* exception and falls back to `"unknown"`. While this is convenient for offline use, it also masks unexpected bugs (e.g., a typo inside `_fetch_weather`). Consider catching only the expected error types (e.g., `NotImplementedError` or a custom `WeatherFetchError`) and letting other exceptions propagate so they surface during development.

  ```python
  try:
      condition = _fetch_weather(city).lower()
  except NotImplementedError:
      condition = "unknown"
  ```

- **Input validation** – The CLI accepts any string for `city`. If a real API is integrated later, ensure the city name is sanitized to avoid injection attacks on the downstream service.

- **Package naming** – The folder hierarchy contains hyphens (`nightly-nightly-emoji-forecast`, `nightly-emoji-forecast`). Hyphens are illegal in Python import paths, which means `python -m utils.nightly_emoji_forecast.src.forecast` will fail. Either:
  1. Rename directories to use underscores (`nightly_emoji_forecast`) **or**
  2. Add a top‑level `__init__.py` that re‑exports the module under a valid name.

- **README command mismatch** – The usage example references `utils.nightly_emoji_forecast.src.forecast`, but the actual path on disk is `utils/nightly-nightly-emoji-forecast/utils/nightly-emoji-forecast/src/forecast.py`. Align the documentation with the real import path or, better, provide a console‑script entry point in `setup.cfg`/`pyproject.toml`:
  ```toml
  [project.scripts]
  nightly-emoji-forecast = "utils.nightly_emoji_forecast.src.forecast:main"
  ```

- **Testing instructions** – The command `python -m unittest discover utils/nightly-emoji-forecast/tests` will not work because the hyphenated directory is not a package. After fixing the naming, the command can stay, or you can add a `pytest` configuration for a smoother experience.

- **Type hints** – `get_emoji_forecast` could be annotated to return `str` explicitly:
  ```python
  def get_emoji_forecast(city: str) -> str:
  ```

## 🧱 Mocks / Fakes
- **Stub design** – `_fetch_weather` raises `NotImplementedError`, which is a clear contract for developers writing their own implementation.  
- **Testability** – The design encourages patching `_fetch_weather` in tests, which is good. Once the import path issue is resolved, the mock will be straightforward:
  ```python
  with patch('utils.nightly_emoji_forecast.src.forecast._fetch_weather', return_value="Snow"):
      assert forecast.get_emoji_forecast("Oslo") == "❄️"
  ```

- **Future extension** – If you later add a real API client, consider extracting an interface (e.g., `class WeatherProvider: def get_condition(self, city: str) -> str`) and inject it into `get_emoji_forecast`. This would let tests replace the provider with a lightweight fake without relying on `patch`.

---

**Overall impression:** The utility is a solid, well‑documented starter. Address the import‑path / hyphen issue, fix the test patching, and tighten exception handling, and the component will be ready for production use and easy for contributors to extend.

PR #1220, opened 2025-11-26

polsala:
## Rationale

## Why safe to merge

## Test Plan

## Links

## Mock Justification

### DIFF
--- a/utils/nightly-nightly-whisperwind-encrypto/utils/nightly-whisperwind-encryptor/README.md
+++ b/utils/nightly-nightly-whisperwind-encrypto/utils/nightly-whisperwind-encryptor/README.md
@@ -1,0 +1,57 @@
+# Nightly Whisperwind Encryptor
+
+A simple, self-contained Python utility for obfuscating and revealing text messages using a Vigenere cipher. In the uncertain times ahead, sometimes a little obscurity is all you need to keep your whispers private.
+
+## Purpose
+
+This utility provides a basic command-line interface to encrypt or decrypt text using a shared passphrase. It's not designed for military-grade security, but rather for quick, offline obfuscation of messages, perfect for passing notes in a post-apocalyptic landscape or simply adding a touch of mystery to your digital communications.
+
+## How it Works
+
+The Whisperwind Encryptor uses the **Vigenere cipher**, a method of encrypting alphabetic text by using a series of different Caesar ciphers based on the letters of a keyword. Non-alphabetic characters (numbers, symbols, spaces) are passed through unchanged, and the original casing of alphabetic characters is preserved.
+
+## Usage
+
+### Prerequisites
+
+*   Python 3.11 (or compatible)
+
+### Running the Utility
+
+Navigate to the `utils/nightly-whisperwind-encryptor/` directory.
+
+#### Encrypting a Message
+
+```bash
+python src/encryptor.py --mode encrypt --text "The quick brown fox jumps over the lazy dog." --key "APOCALYPSE"
+```
+
+Example Output:
+```
+Whisperwind Encrypted: Tpe qeick brmwn fpx jupms ovfr the lzzy dog.
+```
+
+#### Decrypting a Message
+
+```bash
+python src/encryptor.py --mode decrypt --text "Tpe qeick brmwn fpx jupms ovfr the lzzy dog." --key "APOCALYPSE"
+```
+
+Example Output:
+```
+Whisperwind Decrypted: The quick brown fox jumps over the lazy dog.
+```
+
+### Key Considerations
+
+*   The key (passphrase) should ideally be a word or phrase containing only alphabetic characters. Non-alphabetic characters in the key will be ignored.
+*   A key with no alphabetic characters will result in an error.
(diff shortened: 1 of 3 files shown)


polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **Self‑contained utility** – All source, tests, and documentation live under a single `utils/nightly-whisperwind-encryptor` folder, making the component easy to discover and reason about.  
- **Clear separation of concerns** – The core Vigenère logic lives in `vigenere_transform` with tiny helper functions (`_char_to_int`, `_int_to_char`). This keeps the algorithm easy to read and unit‑test.  
- **Graceful error handling** – Invalid keys raise a `ValueError` that is caught in `main()` and reported on `stderr` with a non‑zero exit status.  
- **Case‑preserving encryption** – The implementation respects the original case of alphabetic characters, which is a nice usability touch for a casual obfuscation tool.  
- **CLI ergonomics** – Argument parsing uses `argparse` with explicit `choices` for `--mode`, required flags, and helpful description strings.  
- **Comprehensive test suite** – Tests cover basic encryption/decryption, round‑trip, punctuation handling, case preservation, empty strings, and malformed keys.  

## 🧪 Tests
- **Coverage is good but can be tightened**  
  - The test `test_vigenere_case_preservation` hard‑codes the expected ciphertext (`"NxujtwxbaGP"`). If the algorithm ever changes (e.g., to support Unicode), the test will break even though the behavior may still be correct. Consider generating the expected value with a known‑good reference implementation or documenting the derivation.  
  - The `test_main_*` tests mock `argparse.ArgumentParser.parse_args`. While this works, it also bypasses the parser’s built‑in validation (e.g., missing required arguments). Adding a small integration test that invokes the script via `subprocess.run([...])` would give confidence that the CLI surface behaves as advertised.  

- **Minor test improvements**  
  ```python
  # Example of a more robust case‑preservation test
  def test_vigenere_case_preservation(self):
      message = "ApocalypsAI"
      key = "NIGHTLY"
      encrypted = vigenere_encrypt(message, key)
      # Verify length and case pattern instead of a fixed string
      self.assertEqual(len(encrypted), len(message))
      self.assertTrue(all(c.isupper() if m.isupper() else c.islower()
                          for c, m in zip(encrypted, message)))
  ```  

- **Test naming** – Prefixing all test methods with `test_` is already done, but grouping related tests into sub‑classes (e.g., `class TestEncryption`, `class TestCLI`) can improve readability when the suite grows.  

## 🔒 Security
- **Explicit disclaimer** – The README correctly states that this is *not* for high‑security use. That mitigates misuse expectations.  
- **Key sanitisation** – Non‑alphabetic characters are stripped from the key, which is fine for a toy cipher. However, the current implementation silently drops them; consider warning the user when the supplied key loses characters (e.g., `"K-E-Y!"` becomes `"KEY"`).  
  ```python
  if len(key_upper_alpha) != len(key):
      print("Warning: Non‑alphabetic characters in the key were ignored.", file=sys.stderr)
  ```  
- **Potential side‑channel** – The algorithm runs in pure Python and processes the entire input in memory. For very large texts this could cause high memory usage. Adding a streaming mode (reading line‑by‑line) isn’t required now but could be a future enhancement.  

## 🧩 Docs/DX
- **README completeness** – The README includes purpose, usage examples, and a test command, which is excellent for discoverability.  
- **CLI usage examples** – The examples use hard‑coded text strings. Adding a short “pipe” example would illustrate how the tool can be used in a Unix pipeline:  
  ```bash
  echo "Secret Message" | python src/encryptor.py --mode encrypt --key APOCALYPSE
  ```  
- **Error messages** – The README mentions that a key with no alphabetic characters will error, but it doesn’t show the exact error output. Adding a snippet like:  
  ```
  $ python src/encryptor.py --mode encrypt --text "hi" --key "123"
  Error: Key must contain at least one alphabetic character.
  ```  
  would help users anticipate failure modes.  

- **Versioning / License** – Consider adding a small `VERSION` file or a `__version__` attribute in `encryptor.py`, and include an open‑source license header (e.g., MIT) to clarify reuse rights.  

## 🧱 Mocks/Fakes
- **CLI tests use appropriate mocking** – Patching `argparse.ArgumentParser.parse_args` and `sys.stdout`/`sys.stderr` isolates the `main()` function nicely.  
- **Avoid over‑mocking** – The `test_main_error_handling` also patches `sys.exit`. While this prevents the test runner from exiting, it also hides the fact that `main()` calls `sys.exit(1)`. The current assertion (`mock_exit.assert_called_once_with(1)`) covers this, but you could also verify that the function returns after the error path, ensuring no further processing occurs.  

- **Potential improvement** – Instead of mocking `parse_args`, you could instantiate an `ArgumentParser` with a custom list of arguments using `parser.parse_args(['--mode', 'encrypt', ...])`. This would test the parser’s own validation logic without needing a full subprocess.  

---

- Add a warning for keys that lose characters after sanitisation.  
- Refactor the case‑preservation test to avoid brittle hard‑coded ciphertext.  
- Consider a small integration test that runs the script via `subprocess` to validate the full CLI flow.  
- Expand the README with pipeline usage, explicit error output examples, and a license header.  
- Optionally expose a `__version__` constant for better traceability.  

These tweaks will tighten the utility’s reliability, improve developer experience, and make the documentation a one‑stop shop for both casual users and future maintainers.

PR #1323, opened 2025-11-27

polsala:
## Rationale

## Why safe to merge

## Test Plan

## Links

## Mock Justification

### DIFF
--- a/utils/nightly-nightly-emoji-mood-analyzer/utils/nightly-emoji-mood-analyzer/README.md
+++ b/utils/nightly-nightly-emoji-mood-analyzer/utils/nightly-emoji-mood-analyzer/README.md
@@ -1,0 +1,68 @@
+# Nightly Emoji Mood Analyzer
+
+**What it does**
+
+`emoji-mood-analyzer` scans a string for known emojis and returns a JSON‑compatible dictionary mapping each detected mood to the number of times it appears. It’s perfect for:
+
+- Summarising the emotional tone of a Slack/Discord message.
+- Adding a playful sentiment badge to a commit message.
+- Any situation where you have a handful of emojis and need a quick, deterministic mood breakdown.
+
+**Supported emojis & moods**
+
+| Emoji | Mood |
+|-------|------|
+| 😀, 😃 | happy |
+| 😂 | joy |
+| 😢, 😭 | sad |
+| 😡 | angry |
+| 🤔 | thinking |
+| 👍 | approval |
+| 👎 | disapproval |
+| ❤️ | love |
+| 💔 | heartbreak |
+| 🤯 | mindblown |
+| 🥳 | celebration |
+
+**Installation**
+
+The utility is self‑contained; just copy the `src/` folder into your project or run it directly:
+
+```bash
+python -m utils/nightly-emoji-mood-analyzer/src/analyzer "😀😂👍"
+```
+
+**CLI usage**
+
+```bash
+python -m utils/nightly-emoji-mood-analyzer/src/analyzer "Your emoji string here"
+```
+
+The script prints a pretty‑printed JSON object, e.g.:
+
+```json
+{
+  "happy": 1,
+  "joy": 1,
+  "approval": 1
+}
+```
+
+**Programmatic usage**
+
+```python
+from utils.nightly-emoji-mood-analyzer.src.analyzer import analyze_emojis
+mood_counts = analyze_emojis("😀😂👍")
+```
+
+**Testing**
+
+Run the bundled tests with:
+
+```bash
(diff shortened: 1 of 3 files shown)


polsala:
## ✅ What’s solid
- **Clear purpose & isolation** – The analyzer lives in its own `utils/...` folder and has no external dependencies, making it easy to reason about and to drop into other projects.  
- **Pure functional core** – `analyze_emojis` is side‑effect‑free and returns a plain `dict`; this makes the logic trivially testable and reusable.  
- **Typed mapping** – `EMOJI_MOOD_MAP` is annotated with `Dict[str, str]`, and the helper functions use proper type hints, which helps static analysis tools.  
- **CLI entry‑point** – The `if __name__ == "__main__": _cli()` guard provides a convenient command‑line interface for quick ad‑hoc checks.  
- **Test coverage** – The bundled `unittest` suite checks empty input, unknown characters, single‑mood aggregation, multi‑mood aggregation, and mixed known/unknown characters.  

## 🧪 Tests
- **Positive points**
  - Tests cover the main functional paths and assert exact dictionaries, which is great for deterministic output.
  - The naming (`TestEmojiMoodAnalyzer`) follows the conventional pattern, making the suite discoverable by `unittest`.

- **Opportunities for improvement**
  - **Edge‑case stress test** – Add a test that feeds a very long string (e.g., 10 k emojis) to ensure performance stays acceptable and no overflow occurs.
  - **Non‑string input handling** – Currently the function will raise a `TypeError` if a non‑string is passed. A defensive test (and possibly a guard) would make the API more robust:
    ```python
    def test_non_string_input(self):
        with self.assertRaises(TypeError):
            analyze_emojis(12345)   # or None
    ```
  - **CLI integration test** – Use `subprocess.run` to invoke the module as a script and assert the JSON output. This validates the `-m` entry point and the usage message.
  - **Multi‑codepoint emojis** – Even if the utility intentionally ignores them, a test confirming that a known multi‑codepoint emoji (e.g., "👨‍👩‍👧‍👦") yields an empty result would document the current limitation.

## 🔒 Security
- **Input sanitisation** – The analyzer only reads the supplied string and performs a simple membership check against a hard‑coded map. No external I/O, network calls, or file writes occur, so the attack surface is minimal.
- **JSON dumping** – `json.dumps(..., ensure_ascii=False)` is safe for the data we generate (plain strings and integers). No user‑controlled keys are introduced, so there’s no risk of injection.
- **CLI argument handling** – The script exits with `sys.exit(1)` on incorrect usage, which is fine. Consider adding a `try/except` around the `analyze_emojis` call to catch unexpected exceptions and emit a clean error message rather than a full traceback.

- **Import path typo** – The README shows:
  ```python
  from utils.nightly-emoji-mood-analyzer.src.analyzer import analyze_emojis
  ```
  Hyphens are illegal in Python module names, so this import will fail. Either:
  1. Rename the folder to use underscores (`nightly_emoji_mood_analyzer`) **and** add an `__init__.py` file, **or**
  2. Expose the function via a top‑level package (e.g., create `utils/nightly_emoji_mood_analyzer/__init__.py` that re‑exports `analyze_emojis`).

- **Missing `__init__.py`** – Without an `__init__.py`, the `utils/...` directory isn’t a package, preventing the import shown above. Adding an empty `__init__.py` (or one that defines `__all__ = ["analyze_emojis"]`) resolves this.

- **CLI invocation syntax** – The README suggests:
  ```bash
  python -m utils/nightly-emoji-mood-analyzer/src/analyzer "😀😂👍"
  ```
  The `-m` flag expects a dotted module name, not a file path. After fixing the package name, the correct command would be:
  ```bash
  python -m utils.nightly_emoji_mood_analyzer.src.analyzer "😀😂👍"
  ```
  Update the docs accordingly.

- **README formatting** – The table of emojis is clear, but consider adding a “Limitations” section that mentions:
  - Only single‑codepoint emojis are recognised.
  - No support for skin‑tone modifiers or composite emojis.
  This sets correct expectations for downstream users.

- **Versioning / Distribution** – If the utility is intended for reuse across projects, think about publishing it as a small pip‑installable package (e.g., `nightly-emoji-mood-analyzer`) with a `setup.cfg`/`pyproject.toml`. Even a simple `requirements.txt` entry would make integration smoother.

## 🧱 Mocks / Fakes
- **Not applicable** – The change does not introduce external services or complex dependencies that would require mocking. The current test suite directly exercises the pure functions, which is appropriate.

---

**Overall** – The utility is a tidy, well‑tested addition that delivers immediate value for emoji‑based sentiment extraction. Addressing the import‑path naming and adding a couple of edge‑case tests will make the package more robust and developer‑friendly.

PR #1629, opened 2025-12-04
ato merge to merge

polsala:


polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **General structure** – The file continues to follow the existing pattern of thin wrappers around GitHub API calls, keeping the public surface small and easy to reason about.  
- **Error handling** – The new `merge_pr` function raises a custom `GitHubError` for missing token, HTTP failures, and GraphQL‑level errors, which is consistent with the rest of the module.  
- **Use of environment variable** – Pulling the token from `GITHUB_TOKEN` mirrors the existing authentication strategy, avoiding hard‑coded credentials.  

## 🧪 Tests
- **Add unit tests for the new GraphQL flow**  
  ```python
  @patch.dict(os.environ, {"GITHUB_TOKEN": "test-token"})
  @patch("agents.agent_utils.requests.post")
  @patch("agents.agent_utils.get_pr")
  def test_merge_pr_success(mock_get_pr, mock_post):
      mock_get_pr.return_value = {"node_id": "MDExOlB1bGxSZXF1ZXN0MQ=="}
      mock_post.return_value = Mock(
          status_code=200,
          json=lambda: {
              "data": {
                  "mergePullRequest": {
                      "pullRequest": {"merged": True, "id": "MDExOlB1bGxSZXF1ZXN0MQ=="}
                  }
              }
          },
      )
      merge_pr("owner/repo", 42, merge_method="squash")
      mock_post.assert_called_once()
  ```

- **Test failure paths** – Verify that:
  - Missing `GITHUB_TOKEN` raises `GitHubError`.
  - Non‑2xx HTTP status raises `GitHubError`.
  - GraphQL response containing `"errors"` raises `GitHubError`.
  - `merged` flag is `False` triggers the appropriate exception.

- **Regression test for the original `merge_pr`** – Since the file now defines **two** `merge_pr` functions (the original REST‑based one and the new GraphQL‑based one), add a test that confirms the public API still points to the intended implementation. If the rename was accidental, consider renaming the GraphQL version (e.g., `merge_pr_now`) to avoid shadowing.

## 🔒 Security
- **Validate `merge_method`** – Upper‑casing user input is not enough; an invalid string will be sent to GitHub and cause a 400 error. Add explicit validation:
  ```python
  VALID_METHODS = {"MERGE", "SQUASH", "REBASE"}
  if merge_method.upper() not in VALID_METHODS:
      raise ValueError(f"merge_method must be one of {VALID_METHODS}")
  ```
- **Avoid leaking the token** – The current code never logs the token, which is good. Ensure any future logging of the request/response body redacts the `Authorization` header.  
- **Timeout handling** – A 30‑second timeout is reasonable, but consider making it configurable via a constant or env var for CI environments.

## 🧩 Docs/DX
- **Docstrings** –  
  - Update the docstring of `enable_auto_merge_old` to clearly state that it is deprecated and kept for backward compatibility.  
  - Add a docstring for the new `merge_pr` that distinguishes it from the original “immediate merge” behavior.  
  - Document the expected shape of the `repo` argument (`"owner/name"`), and note that the function will raise `GitHubError` for missing token or API failures.

- **Changelog entry** – Since the public behavior of `enable_auto_merge` now points to `merge_pr`, add a note in the project changelog explaining the breaking change (or revert the alias if it was unintended).

- **Usage example** – Provide a short snippet in the module README or docstring:
  ```python
  # Merge PR #123 immediately using squash
  from agents.agent_utils import merge_pr
  merge_pr("myorg/myrepo", 123, merge_method="squash")
  ```

## 🧱 Mocks/Fakes
- **Mocking `requests.post`** – When writing tests, replace `requests.post` with a mock that returns a `MockResponse` object exposing `status_code`, `text`, and a `json()` method. Example helper:
  ```python
  class MockResponse:
      def __init__(self, status_code=200, json_data=None, text=""):
          self.status_code = status_code
          self._json = json_data or {}
          self.text = text

      def json(self):
          return self._json
  ```

- **Mocking environment** – Use `patch.dict(os.environ, {...})` to inject a fake `GITHUB_TOKEN` for the duration of the test.

- **Mocking `get_pr`** – Since `merge_pr` depends on `get_pr` to fetch the node ID, patch it to return a deterministic payload (`{"node_id": "MDExOlB1bGxSZXF1ZXN0MQ=="}`) rather than hitting the real GitHub API.

- **Avoid double‑definition confusion** – If the original `merge_pr` (REST version) is still needed elsewhere, expose it under a distinct name (e.g., `_merge_pr_rest`) and adjust imports accordingly. This makes mocking clearer and prevents accidental test failures due to function shadowing.

PR #1663, opened 2025-12-05
Revert "Fix bash parsing error when metadata JSON contains single quotes"

polsala:
Reverts polsala/ApocalypsAI#1661


polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **Workflow scaffolding** – The three generation workflows (`gen_gemini.yml`, `gen_groq.yml`, `gen_openrouter.yml`) share a clean, well‑structured layout (checkout, setup, run agent, parse metadata, commit). This makes future maintenance straightforward.  
- **JSON extraction via Python** – Using a tiny Python one‑liner to parse the JSON metadata is robust and avoids brittle `jq` or `sed` tricks. The fallback to a default value (`|| echo "unknown"`) is a nice safety net.  
- **Consistent naming** – The variables `util_name`, `summary`, and `provider` are extracted in the same order across all three workflows, which reduces the chance of copy‑paste errors.  

## 🧪 Tests
- **Add workflow‑level smoke tests** – GitHub Actions can be exercised with the `act` tool or a dedicated CI job that runs the workflow with a mocked `run_agent` step. Verify that:
  ```yaml
  steps:
    - name: Mock run_agent
      id: run_agent
      run: |
        echo "::set-output name=metadata::{\"util_name\":\"test_util\",\"summary\":\"test\",\"provider\":\"gemini\"}"
  ```
  and that the downstream parsing produces the expected `util_name`, `summary`, and `provider` values.
- **Unit‑test the JSON parsing snippet** – Extract the Python command into a small helper script (e.g., `scripts/parse_meta.py`) and add a Python test suite that feeds it various JSON payloads, including:
  - Valid JSON with single quotes inside string values.
  - Empty or malformed JSON to ensure the fallback works.
- **Regression test for the original bug** – Write a test case where the metadata JSON contains a single‑quote character (`'`) inside a string field. The test should assert that the workflow **does not** break when the metadata is quoted with double quotes (the original fix) and that the current revert re‑introduces the failure. This will surface the regression early.

## 🔒 Security
- **Shell quoting** – Switching from `metadata="${{ steps.run_agent.outputs.metadata }}"` to `metadata='${{ steps.run_agent.outputs.metadata }}'` re‑introduces a classic Bash parsing issue: any single‑quote inside the JSON payload will terminate the string early, potentially causing command injection or malformed JSON handling.  
  - **Actionable fix**: Keep the double‑quoted form and, if single‑quotes are a concern, escape them safely:
    ```bash
    metadata="${{ steps.run_agent.outputs.metadata//\'/\\\' }}"
    ```
    or, better, avoid shell interpolation altogether by passing the value via `printf '%s'`:
    ```bash
    metadata=$(printf '%s' "${{ steps.run_agent.outputs.metadata }}")
    ```
- **Output sanitisation** – The `run_agent` step likely runs untrusted LLM output. Consider validating that the `metadata` string is valid JSON before feeding it to Python:
  ```bash
  if ! echo "$metadata" | python3 -c "import sys, json; json.load(sys.stdin)" >/dev/null 2>&1; then
    echo "❌ Invalid metadata JSON"
    exit 1
  fi
  ```
- **Least‑privilege token usage** – Ensure the `GITHUB_TOKEN` used for the commit step has the minimal scopes required (e.g., `contents:write` only). This limits impact if the workflow is ever compromised.

## 🧩 Docs/DX
- **Document the metadata contract** – Add a section in the repository’s `README` or a dedicated `docs/metadata.md` that describes the expected JSON schema (`util_name`, `summary`, `provider`, optional fields) and the quoting rules. Include examples with edge‑case characters (single quotes, newlines, Unicode).  
- **Explain the quoting rationale** – Since the PR reverts a change that fixed a Bash parsing bug, a comment in each workflow file explaining *why* the double‑quoted form is preferred would help future contributors avoid re‑introducing the bug. Example:
  ```yaml
  # NOTE: We use double quotes around the metadata output to allow single‑quote characters
  # inside the JSON payload. See docs/metadata.md for details.
  metadata="${{ steps.run_agent.outputs.metadata }}"
  ```
- **Add a CONTRIBUTING checklist** – Include a step that reminds contributors to run the workflow smoke test locally (via `act`) and to verify JSON quoting before merging.

## 🧱 Mocks/Fakes
- **Mock `run_agent` output in CI** – The current workflows rely on a real LLM call, which makes them flaky and slow. Introduce a conditional `if: ${{ inputs.dry_run }}` or an environment variable (`FAKE_AGENT=true`) that swaps the real step with a deterministic mock:
  ```yaml
  - name: Run agent (or mock)
    id: run_agent
    if: env.FAKE_AGENT == 'true'
    run: |
      echo "::set-output name=metadata::{\"util_name\":\"mock_util\",\"summary\":\"mock summary\",\"provider\":\"gemini\"}"
    else
    uses: ./actions/run-agent
  ```
- **Provide a reusable mock action** – Create a tiny composite action (e.g., `.github/actions/mock-metadata`) that can be referenced across all three workflows. This reduces duplication and ensures the mock output stays in sync with the documented schema.  

---

**Takeaway:** The workflow structure and JSON parsing approach are solid, but reverting the quoting change re‑opens a known Bash parsing vulnerability. Reinstate the safe double‑quoted interpolation (or an explicit escape strategy), add tests that capture the edge case, tighten validation, and improve documentation around the metadata format. These steps will make the CI pipelines more reliable, secure, and developer‑friendly.

polsala:
## 🤖 Review by GROQ Agent

**✅ What's solid**  
- The diff is clean and consistent across all three workflow files.  
- The change is isolated to a single line in each file, making it easy to review and revert if needed.  
- The intent (to revert a previous quoting change) is clearly expressed in the PR title and body.  

**🧪 Tests**  
- No new tests were added.  
- Consider adding a lightweight unit test that verifies the `metadata` variable is parsed correctly when the agent output contains single quotes.  
  ```bash
  # test_metadata_parsing.sh
  metadata='{"util_name":"test","summary":"single quote '\'' test"}'
  util_name=$(echo "$metadata" | python3 -c "import sys,json;print(json.load(sys.stdin).get('util_name'))")
  [ "$util_name" = "test" ] || exit 1
  ```  
- If you have a CI workflow that runs the GitHub Actions locally (e.g., `act`), add a job that feeds a mock agent output to the workflow and asserts the parsed values.  

**🔒 Security**  
- The `metadata` variable is injected directly into a shell command.  
- Using double quotes (`"${{ steps.run_agent.outputs.metadata }}"`) can expose the content to shell interpretation if the JSON contains characters like `$`, `` ` ``, or `;`.  
- The current change re‑introduces the original bug that the single‑quoted version was meant to avoid.  
- A safer approach is to keep the single quotes **or** escape problematic characters before assignment.  
  ```bash
  metadata=$(printf '%s' "${{ steps.run_agent.outputs.metadata }}" | sed "s/'/'\\\\''/g")
  ```  
- Alternatively, avoid shell parsing altogether by passing the JSON directly to the Python one‑liners without intermediate shell variables.  

**🧩 Docs/DX**  
- Add a comment in each workflow explaining why the variable is quoted the way it is.  
  ```yaml
  # NOTE: Use single quotes to preserve JSON containing single quotes
  metadata='${{ steps.run_agent.outputs.metadata }}'
  ```  
- Update any README or internal docs that reference the metadata parsing logic to reflect the current quoting strategy.  
- If the project has a `CONTRIBUTING.md`, include a note about the importance of quoting when dealing with JSON in GitHub Actions.  

**🧱 Mocks/Fakes**  
- Create a mock `run_agent` step that outputs a sample JSON containing single quotes.  
  ```yaml
  - name: Mock run_agent
    id: run_agent
    run: |
      echo "::set-output name=metadata::{\"util_name\":\"test\",\"summary\":\"single quote '\\'' test\"}"
  ```  
- Use this mock in a separate test workflow to ensure the parsing logic still works after the revert.  
- If you use a testing framework for GitHub Actions (e.g., `act` or `github-actions-toolkit`), include this mock as part of the test harness.  

---

**Actionable take‑aways**  
- Re‑evaluate whether the revert is necessary; the original single‑quoted version was a defensive fix.  
- If you decide to keep double quotes, add escaping logic or switch to a safer parsing method.  
- Add at least one unit test or mock workflow to guard against future regressions.  
- Document the reasoning behind the quoting choice so future contributors understand the trade‑offs.

PR #4127, opened 2026-04-09
feat(gemini): A bash script to filter and process syslog messages based on configurable patterns.

polsala:
## Implementation Summary
- **Utility**: nightly-syslog-filter
- **Provider**: gemini
- **Location**: `bash-utils/nightly-nightly-syslog-filter`
- **Files Created**: 3
- **Description**: A bash script to filter and process syslog messages based on configurable patterns.

## Rationale

## Why safe to merge
- Utility is isolated to `bash-utils/nightly-nightly-syslog-filter`.

## Test Plan
- Follow the instructions in the generated README at `bash-utils/nightly-nightly-syslog-filter/README.md`
- Run tests located in `bash-utils/nightly-nightly-syslog-filter/tests/`

## Links

## Mock Justification

### DIFF
--- a/bash-utils/nightly-nightly-syslog-filter/README.md
+++ b/bash-utils/nightly-nightly-syslog-filter/README.md
@@ -1,0 +1,69 @@
+# Nightly Syslog Filter
+
+A whimsical yet useful bash utility to filter and process syslog messages based on configurable patterns. This script helps you keep your logs tidy and highlights important events in the chaos of system messages.
+
+## Philosophy
+
+"Anarchy with discipline" — this script provides a simple, yet powerful, way to manage your logs without complex dependencies. It's designed to be run as a cron job or integrated into other monitoring systems.
+
+## Usage
+
+```bash
+./src/syslog_filter.sh <config_file>
+```
+
+Where `<config_file>` is a path to a configuration file defining the filtering rules.
+
+## Configuration File Format
+
+The configuration file is a simple text file where each line represents a filtering rule. Lines starting with `#` are comments.
+
+Each rule follows the format:
+
+`LEVEL:PATTERN:ACTION`
+
+*   **LEVEL**: The syslog level to match (e.g., `INFO`, `WARNING`, `ERROR`, `CRITICAL`, `DEBUG`). If omitted, all levels are considered.
+*   **PATTERN**: A regular expression to match against the syslog message content.
+*   **ACTION**: What to do with the matched message. Supported actions:
+    *   `LOG`: Print the message to standard output (default if omitted).
+    *   `DROP`: Discard the message.
+    *   `ALERT`: Print the message and send a simple alert (e.g., to stderr).
+
+**Example Configuration (`config.txt`):**
+
+```
+# Ignore all debug messages
+DEBUG:.*:DROP
+
+# Alert on critical errors
+CRITICAL:.*:ALERT
+
+# Log all warnings
+WARNING:.*:LOG
+
+# Log specific kernel messages
+:kernel:.*:LOG
+
+# Drop noisy application messages
+INFO:noisy_app:.*:DROP
+```
+
+## Examples
+
+1.  **Basic Filtering**: Filter logs using `config.txt` and output to a file.
+    ```bash
+    ./src/syslog_filter.sh config.txt > filtered_logs.txt
+    ```
+
+2.  **Alerting Only**: Show only critical alerts.
+    ```bash
+    ./src/syslog_filter.sh config.txt | grep "ALERT:"
+    ```
+
+## Testing
+
+This utility comes with a set of deterministic tests that do not require actual syslog access. You can run them using `bash`.
+
+```bash
+./tests/test_syslog_filter.sh
+```
--- a/bash-utils/nightly-nightly-syslog-filter/src/syslog_filter.sh
+++ b/bash-utils/nightly-nightly-syslog-filter/src/syslog_filter.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+
+# Nightly Syslog Filter
+# Filters syslog messages based on a configuration file.
+
+# --- Configuration --- 
+# Default configuration file if none is provided
+DEFAULT_CONFIG="config.txt"
+
+# --- Helper Functions ---
+
+# Function to log messages with a specific prefix
+log_message() {
+    local prefix="$1"
+    local message="$2"
+    echo "$prefix: $message"
+}
+
+# Function to send an alert
+alert_message() {
+    local level="$1"
+    local message="$2"
+    log_message "ALERT ($level)" "$message"
+    # In a real-world scenario, this could send an email, trigger a webhook, etc.
+    # For this utility, we'll just echo to stderr.
+    echo "ALERT: Critical event detected - $message" >&2
+}
+
+# --- Main Logic ---
+
+# Determine the configuration file
+CONFIG_FILE=${1:-$DEFAULT_CONFIG}
+
+# Check if configuration file exists
+if [ ! -f "$CONFIG_FILE" ]; then
+    echo "Error: Configuration file '$CONFIG_FILE' not found." >&2
+    exit 1
+fi
+
+# Read and process the configuration file line by line
+while IFS= read -r line || [[ -n "$line" ]]; do
+    # Skip comments and empty lines
+    if [[ "$line" =~ ^#.* ]] || [ -z "$line" ]; then
+        continue
+    fi
+
+    # Parse the rule: LEVEL:PATTERN:ACTION
+    IFS=':' read -r LEVEL PATTERN ACTION <<< "$line"
+
+    # Default action is LOG if not specified
+    ACTION=${ACTION:-LOG}
+
+    # Construct the grep pattern, handling optional level
+    GREP_PATTERN=""
+    if [ -n "$LEVEL" ]; then
+        # Basic syslog level matching (can be extended for more complex formats)
+        GREP_PATTERN="^$LEVEL:"
+    fi
+
+    if [ -n "$PATTERN" ]; then
+        if [ -n "$GREP_PATTERN" ]; then
+            GREP_PATTERN="$GREP_PATTERN.*$PATTERN"
+        else
+            GREP_PATTERN="^.*$PATTERN"
+        fi
+    fi
+
+    # If no pattern is specified, it means match all lines for the given level (or all if no level)
+    if [ -z "$GREP_PATTERN" ] && [ -n "$LEVEL" ]; then
+        GREP_PATTERN="^$LEVEL:"
+    fi
+
+    # Apply the filter to standard input
+    # We use a subshell to process each rule independently
+    ( 
+        if [ -n "$GREP_PATTERN" ]; then
+            grep -E "$GREP_PATTERN"
+        else
+            # If no GREP_PATTERN, it means we should process all lines for the specified level, or all lines if no level specified.
+            # This case is tricky and might need refinement based on exact syslog format. For now, if no pattern, we assume it's a broad rule.
+            # If only LEVEL is specified, we match lines starting with that level.
+            if [ -n "$LEVEL" ]; then
+                grep "^$LEVEL:"
+            else
+                # If neither LEVEL nor PATTERN, this rule is likely malformed or intended to match everything.
+                # For safety, we'll assume it's an error or a no-op if no specific pattern.
+                # A more robust solution would validate this.
+                cat # Pass through all lines if no specific pattern and no level
+            fi
+        fi
+    ) <&0 | while IFS= read -r msg || [[ -n "$msg" ]]; do
+        case "$ACTION" in
+            "DROP")
+                # Do nothing, effectively dropping the message
+                ;; 
+            "ALERT")
+                # Log the original message and send an alert
+                alert_message "$LEVEL" "$msg"
+                ;; 
+            "LOG" | *)
+                # Log the message with its original format
+                echo "$msg"
+                ;; 
+        esac
+    done
+
+done < "$CONFIG_FILE"
+
+exit 0
--- a/bash-utils/nightly-nightly-syslog-filter/tests/test_syslog_filter.sh
+++ b/bash-utils/nightly-nightly-syslog-filter/tests/test_syslog_filter.sh
@@ -1,0 +1,128 @@
+#!/bin/bash
+
+# Nightly Syslog Filter - Unit Tests
+
+# Mocking functions for testing
+
+# Mock rationale: This function simulates sending an alert. In a real scenario, it might send an email or trigger an external system. For testing, we capture its output.
+alert_message() {
+    local level="$1"
+    local message="$2"
+    echo "MOCK_ALERT:$level:$message" >&2
+}
+
+# Mock rationale: This function simulates logging a message. For testing, we capture its output.
+log_message() {
+    local prefix="$1"
+    local message="$2"
+    echo "MOCK_LOG:$prefix:$message"
+}
+
+# --- Test Setup ---
+
+# Source the main script to use its functions and variables (but override mocked ones)
+# Mock rationale: Sourcing the script allows us to test its logic without executing it directly in a live environment. We've already mocked critical functions.
+. "src/syslog_filter.sh"
+
+# Create a temporary directory for test configurations
+TEST_DIR=$(mktemp -d)
+
+# Function to clean up temporary files
+cleanup() {
+    rm -rf "$TEST_DIR"
+}
+
+# Register cleanup function to run on exit
+trap cleanup EXIT
+
+# --- Test Cases ---
+
+run_test() {
+    local test_name="$1"
+    local input_log="$2"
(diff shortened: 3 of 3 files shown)


polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **Clear purpose & isolation** – The utility lives in its own `bash-utils/nightly-nightly-syslog-filter` folder and does not touch any existing code‑base, making it easy to reason about.
- **Self‑documenting README** – The README explains the philosophy, usage, configuration format, and provides concrete examples. New contributors can get started without digging into the script.
- **Deterministic test suite** – Tests are bundled with the utility and can be run with a single command (`./tests/test_syslog_filter.sh`). This is great for CI integration.
- **Graceful handling of missing config** – The script checks for the existence of the configuration file and exits with a clear error message.
- **Extensible rule format** – The `LEVEL:PATTERN:ACTION` syntax is simple yet expressive enough for many common syslog filtering scenarios.

---

## 🧪 Tests
| Tests **source** the script (`. "src/syslog_filter.sh"`). Because the script executes its main logic at load time, sourcing it runs the whole filter immediately (reading from stdin, exiting, etc.). This makes the mocked `log_message` / `alert_message` functions ineffective – they are overwritten by the definitions inside the sourced script. | **Refactor the script** so that the core logic lives in functions (e.g., `process_logs`) and the top‑level only parses arguments and calls those functions. Then the test harness can source the file, replace the helper functions, and invoke `process_logs` directly. |
| The test harness creates a temporary config file per test, which is good, but it feeds the log data via `echo -e "$input_log"`. This relies on Bash’s `-e` handling of `\n` and can be fragile if the log lines contain backslashes or other escape sequences. | Use a **here‑document** or a temporary file for the input log to avoid escape‑sequence pitfalls, e.g.:<br>`cat <<'EOF' >"$TEST_DIR/input.log"`<br>`$input_log`<br>`EOF`<br>`./src/syslog_filter.sh "$config_file" < "$TEST_DIR/input.log"` |
| Expected stdout/stderr strings are compared with plain string equality. Whitespace differences (trailing newlines) can cause false negatives. | Trim trailing newlines before comparison or use a diff tool (`diff -u <(echo "$expected") <(echo "$actual")`) to provide more helpful diff output when a test fails. |
| Test **7** (missing config) runs the script directly but does not assert the exit status. | Capture `$?` and assert that it is non‑zero (e.g., `if [[ $? -ne 0 && ... ]]; then`). This guarantees the script fails as intended. |
| No test covers **invalid rule syntax** (e.g., missing fields, malformed regex). | Add a test case that supplies a bad rule and verifies that the script either skips it with a warning or exits with an informative error. This will protect against future regressions. |

---

## 🔒 Security
- **Untrusted regex injection** – The configuration file is parsed verbatim and passed to `grep -E`. A malicious pattern could cause catastrophic back‑tracking or even trigger `grep` to read from files if the pattern contains `-f`‑style constructs (unlikely but possible with certain `grep` versions).  
  **Mitigation:** Validate each pattern before using it. A simple sanity check could be: `if ! grep -E -q "$PATTERN" <<<""; then echo "Invalid regex in rule: $line" >&2; continue; fi`. For stricter safety, consider using `awk` or `sed` with explicit quoting, or limit the allowed characters in patterns.

- **Shell‑injection via config** – The script builds `GREP_PATTERN` by concatenating strings without quoting. If a config line contains characters that the shell interprets (e.g., backticks, `$()`) and later gets evaluated in a subshell, it could lead to command execution.  
  **Mitigation:** Use **array** syntax for `grep` arguments to avoid word‑splitting and globbing, e.g., `grep -E -- "$GREP_PATTERN"`.

- **File descriptor handling** – The script reads from `stdin` inside a `while` loop that also reads the config file. Because each rule spawns a subshell that consumes `stdin`, subsequent rules may see an empty stream. This is more of a functional bug, but it can also be a denial‑of‑service vector if a malformed config silently drops all logs.  
  **Mitigation:** Read the entire input into a variable or temporary file once, then apply each rule against that stored data. Example skeleton:

  ```bash
  INPUT=$(cat)               # capture all log lines
  while IFS= read -r line; do
      # build GREP_PATTERN …
      case "$ACTION" in
          DROP)   ;;
          ALERT) echo "$INPUT" | grep -E "$GREP_PATTERN" | while read -r msg; do alert_message "$LEVEL" "$msg"; done ;;
          LOG|*)  echo "$INPUT" | grep -E "$GREP_PATTERN"
      esac
  done < "$CONFIG_FILE"
  ```

- **Exit codes** – The script exits with `0` even when a rule is malformed (it just “passes through”). It would be safer to exit with a non‑zero status on configuration errors, making CI pipelines detect misconfiguration early.

---

- **Installation instructions** – The README assumes the script is executable (`./src/syslog_filter.sh`). Adding a quick “make the script executable” tip (`chmod +x src/syslog_filter.sh`) would help newcomers.
- **Configuration file discovery** – Mention that the default config (`config.txt`) is looked for **relative to the current working directory**, not the script location. This can be a source of confusion when the script is invoked from a different directory.
- **Supported syslog formats** – The script currently matches lines that start with `LEVEL:`. Real syslog lines often include timestamps and hostnames (`<PRI>timestamp hostname TAG: message`). Clarify the expected input format or provide a flag to enable a more robust parser.
- **Exit status documentation** – Document the meaning of exit codes (`0` = success, `1` = missing config, `2` = malformed rule, etc.).
- **Extensibility guide** – Since the utility is meant to be used in cron jobs or pipelines, a short section on how to pipe the output to other tools (e.g., `logger`, `mail`) would improve the developer experience.

---

## 🧱 Mocks / Fakes
- **Current mock strategy is overridden** – The test file defines `alert_message` and `log_message` *before* sourcing the script, but the script re‑defines those functions, wiping out the mocks. Consequently the tests are exercising the real implementations, not the mocked ones.  
  **Fix:** Move the `source` line **after** the mock definitions *and* rename the functions in the script (e.g., `real_alert_message`, `real_log_message`) or source the script into a subshell and export the mock functions with `export -f`. Example:

  ```bash
  # In test file
  alert_message() { echo "MOCK_ALERT:$1:$2" >&2; }
  log_message()   { echo "MOCK_LOG:$1:$2"; }

  # Source the script in a subshell so the mocks stay in effect
  ( . "$PROJECT_ROOT/src/syslog_filter.sh" )
  ```

- **Avoid sourcing the whole script** – As mentioned in the Tests section, refactor the script to expose a single entry‑point function (e.g., `filter_syslog`) that accepts input via a parameter or stdin. Then the test can call that function directly after mocking, without needing to source the script’s top‑level execution block.

- **Mocking `grep` for edge‑case testing** – Some security tests (e.g., regex back‑tracking) could benefit from a mock `grep` that records the pattern it receives. This can be achieved by temporarily overriding `grep` in the test’s `PATH`:

  ```bash
  mkdir -p "$TEST_DIR/bin"
  echo -e '#!/bin/bash\nprintf "GREP_CALLED:%s\n" "$@"' >"$TEST_DIR/bin/grep"
  chmod +x "$TEST_DIR/bin/grep"
  PATH="$TEST_DIR/bin:$PATH" ./src/syslog_filter.sh "$config_file" < "$log_file"
  ```

  This technique lets you assert that the script is constructing the expected pattern without actually running the real `grep`.

---

### TL;DR Action Items
1. **Refactor** the script: move core logic into reusable functions, keep top‑level thin.
2. **Fix test mocks**: ensure mocked `log_message` / `alert_message` survive the source step, or call the core function directly.
3. **Add validation** for configuration lines (missing fields, invalid regex) and exit with a non‑zero status on errors.
4. **Quote/escape** all dynamic values passed to `grep` to prevent shell‑injection and pattern‑related crashes.
5. **Capture and test exit codes** for error paths (missing config, malformed rule).
6. **Enhance README** with installation tip, config file location clarification, supported syslog format notes, and exit‑code documentation.
7. **Add tests** for malformed rules, empty config, and a regression test that ensures multiple rules are applied to the *same* input stream (i.e., the input isn’t consumed after the first rule). 

Implementing these changes will make the utility more robust, secure, and maintainable while preserving the nice developer experience you’ve already built. 🚀

polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **Clear purpose & scope** – The utility lives in its own `bash-utils/nightly-nightly-syslog-filter` folder and does not touch any existing code‑base, making the change low‑risk from an integration standpoint.  
- **Self‑contained documentation** – The README explains the philosophy, usage, configuration format, and provides concrete examples. This is very helpful for future contributors and users.  
- **Test harness included** – A Bash‑only test suite is shipped alongside the script, which is great for a language that often lacks automated testing.  
- **Graceful error handling for missing config** – The script exits with a clear error message when the supplied configuration file cannot be found.  

## 🧪 Tests
| **Mocks are overridden** – The test file defines `alert_message` and `log_message` *before* sourcing `src/syslog_filter.sh`. When the script is sourced it re‑defines those functions, wiping out the mocks. | Move the `source` line **after** the mock definitions, or source the script in a subshell and then `export -f` the mocks so they take precedence. Example: <br>```bash\n# Define mocks\nmock_alert_message() { … }\nmock_log_message() { … }\n# Export them for the subshell that will source the script\nexport -f mock_alert_message mock_log_message\n# Source the script inside a subshell that uses the mocks\n(\n  source src/syslog_filter.sh\n  # run the test logic here\n)\n``` |
| **Expected output mismatches** – Tests expect `MOCK_LOG:` and `MOCK_ALERT:` prefixes, but the real `log_message`/`alert_message` functions emit `LOG:` / `ALERT:`. Because of the overriding issue above, the assertions will always fail. | Align the expectations with the actual output, or better, keep the mock approach and ensure the script uses the mocked functions (see previous point). |
| **Single‑pass vs. per‑rule processing** – The script processes the input stream **once per rule** (`while IFS= read -r line …; … (grep …) <&0 …`). After the first rule consumes `stdin`, subsequent rules receive an empty stream, causing many rules to be ineffective. | Refactor the core loop to read the log **once** and evaluate each line against *all* rules. A simple approach: <br>```bash\nwhile IFS= read -r log_line; do\n  handled=false\n  while IFS= read -r rule; do\n    # parse rule and decide action\n    # if rule matches log_line { …; handled=true; break }\n  done < \"$CONFIG_FILE\"\n  # default LOG if not handled\n  $handled || echo \"$log_line\"\ndone\n``` |
| **Exit status not verified** – Test 7 checks for an error message but does not assert that the script exits with a non‑zero status. | Capture `$?` after invoking the script and assert it is non‑zero: <br>```bash\n./src/syslog_filter.sh missing.cfg 2>stderr.log\nif [[ $? -ne 0 && -s stderr.log ]]; then …; fi\n``` |
| **Test readability** – The `run_test` helper mixes positional arguments and long inline strings, making it hard to see the intent at a glance. | Use associative arrays or named variables for each component (input, config, expected stdout, expected stderr). This improves maintainability and makes future test additions easier. |

## 🔒 Security
- **Untrusted regex injection** – The configuration file feeds directly into `grep -E "$GREP_PATTERN"`. If an attacker can control the config, they could supply a malicious regular expression that triggers catastrophic backtracking (ReDoS) or even cause the script to hang.  
  - *Mitigation*: Validate each pattern before using it. For example, reject patterns longer than a reasonable length or containing known problematic constructs (`.*.*.*`, nested quantifiers, etc.). You could also run `grep -E -m 1` to limit matches per line, or use `awk` with a safer matching model.  
- **Potential command‑injection via `eval`‑like behavior** – While the script does not use `eval`, the use of `IFS=':' read -r LEVEL PATTERN ACTION <<< "$line"` will split on colons. If a rule contains backticks or `$()` in the `PATTERN` field, they will be passed unchanged to `grep`, which treats them as literals, but future modifications that embed the pattern in a command string could become vulnerable.  
  - *Mitigation*: Quote variables rigorously (`grep -E "$GREP_PATTERN"` is already quoted, but ensure any future expansions also use `printf '%s\n' "$var"` rather than direct interpolation).  
- **File permission assumptions** – The script expects the config file to be readable. If the file is world‑writable, an attacker could modify it on the fly.  
  - *Mitigation*: Add a sanity check: `[[ -r "$CONFIG_FILE" && ! -w "$CONFIG_FILE" ]] || echo "Warning: config file is writable by others"`.

- **Missing example config file** – The README mentions a default `config.txt` but the repository does not ship one. Adding a minimal `config.example.txt` (and referencing it in the README) would help users get started quickly.  
- **Executable flag** – Ensure `src/syslog_filter.sh` is committed with the executable bit (`chmod +x`). CI may not preserve it; add a step in the CI pipeline to `chmod +x` before testing, or set `git update-index --chmod=+x`.  
- **Usage of stdin** – The README states `./src/syslog_filter.sh <config_file>` but does not clarify that the script expects log lines on **STDIN**. Adding a note such as “Pipe syslog entries into the script, e.g. `journalctl -f | ./src/syslog_filter.sh config.txt`” would prevent confusion.  
- **Exit codes** – Document the meaning of exit codes (e.g., `0` = success, `1` = missing config, `2` = malformed rule). This aids automation scripts that may rely on the return status.  
- **Versioning / License** – Consider adding a version header or SPDX identifier at the top of the script for easier tracking and compliance.  

## 🧱 Mocks / Fakes
- **Current mock strategy is fragile** – Overriding functions after sourcing the script does not work because Bash redefines them. A more robust pattern is to **source the script inside a subshell** where the mocks are already exported, or to **use `declare -f`** to replace the functions after sourcing. Example: <br>```bash\n# Define mocks\nmock_alert_message() { echo "MOCK_ALERT:$1:$2" >&2; }\nmock_log_message() { echo "MOCK_LOG:$1:$2"; }\nexport -f mock_alert_message mock_log_message\n# Source the script in a subshell that will pick up the exported mocks\n(\n  source src/syslog_filter.sh\n  # run test logic here\n)\n``` |
- **Avoid global state** – The script relies on global variables (`DEFAULT_CONFIG`, `CONFIG_FILE`). In tests, consider resetting these between runs to avoid cross‑test contamination.  
- **Consider a test harness library** – For Bash projects, tools like **Bats** or **shunit2** provide a more structured testing environment (setup/teardown, assertions). Migrating the test suite to one of these would improve readability and reporting.  

---

1. **Refactor the core filtering loop** to process each log line once against all rules (prevents stdin exhaustion).  
2. **Re‑order or isolate mock definitions** so they actually replace the real functions during testing.  
3. **Add validation for configuration entries** (syntax checks, pattern length limits) to mitigate ReDoS and malformed‑rule risks.  
4. **Ship an example `config.example.txt`** and reference it in the README.  
5. **Ensure the script file is executable** in the repository and mention the need for stdin in the usage docs.  
6. **Document exit codes** and consider adding an SPDX license header.  
7. **Optionally migrate tests to Bats/shunit2** for clearer test semantics and better CI integration.  

Implementing these changes will make the utility more reliable, secure, and easier for contributors to adopt and extend.

polsala:

✅ **What's solid**

- **Isolated utility with clear scope**: The `nightly-syslog-filter` utility is well-contained within its own directory and doesn't interfere with existing code, aligning with safe merge practices.
- **Comprehensive documentation**: The included `README.md` clearly explains usage, configuration format, and provides practical examples—great for quick onboarding.
- **Deterministic testing approach**: Tests are self-contained and don’t rely on external syslog access, ensuring reproducibility across environments.

🧪 **Tests**

- **Good coverage of core functionality**: Tests cover basic filtering, pattern matching, alerting, and edge cases like missing config files.
- **Mocking strategy needs clarification**: While mocks are present, they override internal functions (`log_message`, `alert_message`) which may lead to false positives if the real implementation diverges. Consider mocking at the I/O boundary instead:

```bash
# Instead of overriding log_message/alert_message,
# mock the actual side effects such as writing to stderr or network calls.
```

- **Missing negative test scenarios**: Add tests for malformed configuration lines or invalid regex patterns to ensure graceful error handling.

🔒 **Security**

- **No hardcoded secrets or credentials**: As noted in the PR description, there’s no exposure of sensitive data.
- **Input validation improvements needed**: The script currently trusts user-provided regex patterns without sanitization. Malformed patterns could cause runtime failures or unexpected behavior. Consider adding checks:

```bash
# Example enhancement
if ! echo "" | grep -E "$PATTERN" >/dev/null 2>&1; then
  echo "Invalid regex pattern: $PATTERN" >&2
  continue
fi
```

🧩 **Docs/DX**

- **Clear usage instructions and examples**: The README makes it easy for users to understand how to run and configure the tool.
- **Configuration format explanation lacks detail**: Clarify how multi-part fields (like timestamps or structured data) in syslog lines should be handled. Also, document expected input stream format.
- **Add installation/setup notes**: Include prerequisites (e.g., Bash version, required tools like `grep`) and example cron integration:

```markdown
## Installation

Ensure `grep` and `bash` are installed.

To schedule hourly log filtering:
```cron
0 * * * * /path/to/src/syslog_filter.sh /path/to/config.txt >> /var/log/filtered.log
```
```

🧱 **Mocks/Fakes**

- **In-script logic mocking reduces fidelity**: Mocking internal helper functions rather than simulating full execution paths limits test realism. Refactor tests to invoke the script as a subprocess while controlling inputs via stdin and config files.
- **Temporary file management is sound**: Proper use of `mktemp` and cleanup traps ensures tests don’t leave artifacts behind. Maintain this practice for future expansions.

PR #4367, opened 2026-04-27
feat(groq): An Ansible playbook that rotates system logs, compresses old logs, and uploads them to a remote backup location.

polsala:
## Implementation Summary
- **Utility**: nightly-log-rotate-archive
- **Provider**: groq
- **Location**: `ansible-playbooks/nightly-nightly-log-rotate-archive`
- **Files Created**: 3
- **Description**: An Ansible playbook that rotates system logs, compresses old logs, and uploads them to a remote backup location.

## Rationale
- This utility was generated using the **groq** AI provider.

## Why safe to merge
- Utility is isolated to `ansible-playbooks/nightly-nightly-log-rotate-archive`.

## Test Plan
- Follow the instructions in the generated README at `ansible-playbooks/nightly-nightly-log-rotate-archive/README.md`
- Run tests located in `ansible-playbooks/nightly-nightly-log-rotate-archive/tests/`

## Links

## Mock Justification

### DIFF
--- a/ansible-playbooks/nightly-nightly-log-rotate-archive/README.md
+++ b/ansible-playbooks/nightly-nightly-log-rotate-archive/README.md
@@ -1,0 +1,46 @@
+# Nightly Log Rotate & Archive
+
+## Overview
+
+This Ansible playbook rotates the system logs in `/var/log`, compresses logs older than a configurable number of days, and uploads the archived logs to a remote backup server via `rsync`.
+
+## Features
+
+- Safe rotation using `logrotate` configuration.
+- Compression of rotated logs with `gzip`.
+- Optional remote backup via `rsync` (SSH).
+- Idempotent – safe to run repeatedly.
+
+## Requirements
+
+- Ansible 2.9+ installed on the control node.
+- Target hosts must have `logrotate`, `gzip`, and `rsync` available.
+- SSH access to the remote backup server (if backup is enabled).
+
+## Variables
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `logrotate_path` | `/etc/logrotate.d` | Directory where logrotate config files are placed. |
+| `log_files` | `["/var/log/*.log"]` | List of log file glob patterns to rotate. |
+| `rotate_days` | `7` | Keep rotated logs for this many days before deletion. |
+| `backup_enabled` | `false` | Set to `true` to enable remote backup. |
+| `backup_destination` | `"backup@example.com:/backups/logs"` | Remote rsync destination (user@host:/path). |
+| `backup_user` | `"backup"` | SSH user for remote backup. |
+
+## Usage
+
+```bash
+ansible-playbook -i inventory.ini src/log_rotate.yml \
+  -e "logrotate_path=/etc/logrotate.d log_files=['/var/log/syslog','/var/log/auth.log'] rotate_days=14 backup_enabled=true backup_destination='backup@example.com:/backups/logs'"
+```
+
+## Testing
+
+Run the included test playbook with:
+
+```bash
+ansible-playbook -i localhost, tests/test_log_rotate.yml
+```
+
+The test uses mock variables and does not modify real system logs.
--- a/ansible-playbooks/nightly-nightly-log-rotate-archive/src/log_rotate.yml
+++ b/ansible-playbooks/nightly-nightly-log-rotate-archive/src/log_rotate.yml
@@ -1,0 +1,72 @@
+---
+- name: Nightly Log Rotate & Archive
+  hosts: all
+  become: true
+  vars:
+    logrotate_path: "/etc/logrotate.d"
+    log_files:
+      - "/var/log/*.log"
+    rotate_days: 7
+    backup_enabled: false
+    backup_destination: "backup@example.com:/backups/logs"
+    backup_user: "backup"
+  tasks:
+    - name: Ensure logrotate configuration directory exists
+      file:
+        path: "{{ logrotate_path }}"
+        state: directory
+        mode: "0755"
+
+    - name: Deploy logrotate configuration
+      copy:
+        dest: "{{ logrotate_path }}/nightly_logrotate.conf"
+        content: |
+          {{ log_files | join(' ') }} {
+              daily
+              rotate {{ rotate_days }}
+              compress
+              missingok
+              notifempty
+              create 0640 root adm
+              sharedscripts
+              postrotate
+                  /usr/lib/rsyslog/rsyslog-rotate
+              endscript
+          }
+        mode: "0644"
+
+    - name: Force logrotate run (dry-run for safety)
+      command: logrotate -d {{ logrotate_path }}/nightly_logrotate.conf
+      register: logrotate_output
+      changed_when: false
+
+    - name: Debug logrotate output
+      debug:
+        var: logrotate_output.stdout_lines
+
+    - name: Perform actual logrotate if not in check mode
+      command: logrotate {{ logrotate_path }}/nightly_logrotate.conf
+      when: not ansible_check_mode
+
+    - name: Find rotated log files
+      find:
+        paths: "/var/log"
+        patterns: "*.gz"
+        age: "{{ rotate_days }}d"
+        age_stamp: mtime
+      register: rotated_logs
+
+    - name: Ensure backup destination is reachable (when backup enabled)
+      ping:
+      delegate_to: "{{ backup_destination.split(':')[0] }}"
+      when: backup_enabled
+
+    - name: Rsync rotated logs to remote backup
+      synchronize:
+        src: "{{ item.path }}"
+        dest: "{{ backup_destination }}/"
+        mode: push
+        rsync_opts:
+          - "--chmod=0644"
+      loop: "{{ rotated_logs.files }}"
+      when: backup_enabled
--- a/ansible-playbooks/nightly-nightly-log-rotate-archive/tests/test_log_rotate.yml
+++ b/ansible-playbooks/nightly-nightly-log-rotate-archive/tests/test_log_rotate.yml
@@ -1,0 +1,80 @@
+---
+- name: Test Nightly Log Rotate & Archive Playbook
+  hosts: localhost
+  gather_facts: false
+  vars:
+    # Mock variables for testing
+    logrotate_path: "/tmp/mock_logrotate"
+    log_files:
+      - "/tmp/mock_logs/app.log"
+    rotate_days: 3
+    backup_enabled: true
+    backup_destination: "localhost:/tmp/mock_backup"
+    backup_user: "root"
+  tasks:
+    - name: Create mock log directory
+      file:
+        path: "/tmp/mock_logs"
+        state: directory
+        mode: "0755"
+
+    - name: Create a dummy log file
+      copy:
+        dest: "/tmp/mock_logs/app.log"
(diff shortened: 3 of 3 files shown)


polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **Isolation & Scope** – All new files live under `ansible‑playbooks/nightly‑nightly‑log‑rotate‑archive`, so the change is self‑contained and won’t affect unrelated playbooks.  
- **Idempotent design** – The playbook creates the logrotate config directory, copies a static config file, and runs `logrotate` only when not in check mode. This follows Ansible best‑practice for repeatable runs.  
- **Variable‑driven** – All paths, log lists, rotation period, and backup toggles are exposed as variables with sensible defaults, making the utility easy to customise per environment.  
- **Graceful backup gating** – The `backup_enabled` flag gates both the reachability ping and the `synchronize` task, preventing accidental remote calls when the feature is disabled.  
- **README completeness** – The README explains purpose, requirements, variables, and a one‑liner usage example, which is a good starting point for new contributors.  

## 🧪 Tests
| Observation | Suggested improvement |
|-------------|-----------------------|
| The test playbook creates mock directories under `/tmp` and includes the main playbook via `include_tasks: "../src/log_rotate.yml"`. | Use `import_playbook` (or `include_role` if you refactor to a role) to bring in the full playbook context. `include_tasks` runs the file as a task list, which can cause variable scoping issues and may skip top‑level `hosts`/`become` settings. |
| `logrotate` is invoked directly in the main playbook, but the test environment may not have the `logrotate` binary installed. | Add a pre‑task that checks for `logrotate` (e.g., `ansible.builtin.command: which logrotate`) and fails with a clear message if missing. In CI you can install it via the appropriate package manager (`apt-get install -y logrotate` or `yum install -y logrotate`). |
| The test asserts the existence of `app.log.1.gz` immediately after the playbook runs, but `logrotate` may create a file with a different suffix (`.gz` without the `.1` if `rotate` is set to 1). | Adjust the assertion to look for any `*.gz` file in the mock directory, or explicitly set `rotate: 1` in the test variables to guarantee the expected name. |
| `backup_enabled` is set to `true` and `backup_destination` points to `localhost:/tmp/mock_backup`. The `synchronize` module will try to SSH to *localhost* using the default user, which may not be `root`. | Pass the SSH user explicitly: `ssh_user: "{{ backup_user }}"` inside the `synchronize` task, or set `delegate_to: localhost` and `mode: push` with `rsync_path` pointing to a local copy command. |
| The `ping` task uses `delegate_to: "{{ backup_destination.split(':')[0] }}"`. This works for simple `user@host:path` strings but will break if the destination includes a port or uses an IPv6 address. | Extract the host with a more robust Jinja filter, e.g.: `{{ backup_destination | regex_replace('^([^@]+@)?([^:]+):.*$', '\\2') }}` and document the expected format. |
| Cleanup step removes the mock directories **after** the assertions, which is fine, but if a test fails the cleanup never runs, leaving stray files on the runner. | Wrap the cleanup in a `block/rescue/alway` pattern so it executes regardless of earlier failures. |

### Example test tweak
```yaml
- name: Import the full playbook (preserves hosts/become)
  import_playbook: ../src/log_rotate.yml

- name: Verify rotated log exists (any gz file)
  find:
    paths: "/tmp/mock_logs"
    patterns: "*.gz"
  register: rotated

- name: Assert at least one rotated file
  assert:
    that:
      - rotated.matched > 0
    fail_msg: "No rotated/compressed logs were found."
```

## 🔒 Security
- **Privilege escalation** – The playbook runs with `become: true`. Consider adding a `become_user: root` explicitly, and document that the playbook must be executed by a user with sudo rights.  
- **Remote backup credentials** – The `backup_user` variable is defined but never used in the `synchronize` task. This could lead to accidental use of the control‑node’s SSH identity. Add `ssh_user: "{{ backup_user }}"` (or `remote_user`) to the `synchronize` task to enforce the intended account.  
- **Path handling** – `backup_destination` is concatenated directly into the `synchronize` `dest` argument. If a malicious contributor supplies a value like `"; rm -rf /"` it could be interpreted by the underlying shell. Use the `rsync_path` option to avoid shell interpolation, e.g.:  
  ```yaml
  synchronize:
    src: "{{ item.path }}"
    dest: "{{ backup_destination }}/"
    mode: push
    rsync_opts:
      - "--chmod=0644"
    rsync_path: "rsync"
  ```  
- **Input validation** – The playbook trusts the `log_files` list. If a user passes a path like `/etc/passwd`, `logrotate` will attempt to rotate it. Add a simple sanity check:  
  ```yaml
  - name: Validate log file paths are under /var/log
    assert:
      that:
        - item.startswith('/var/log/')
    loop: "{{ log_files }}"
    fail_msg: "Only logs under /var/log may be rotated."
  ```

- **README – Variable table** – The table is clear, but adding the *type* (list, string, bool) and *example* values would help newcomers.  
- **Usage example** – Show both the default run and an example with backup enabled, e.g.:  
  ```bash
  ansible-playbook -i inventory.ini src/log_rotate.yml \
    -e "backup_enabled=true backup_destination='backup@example.com:/backups/logs' rotate_days=14"
  ```  
- **Testing instructions** – Mention that the test requires `logrotate` and `rsync` on the host, and that it should be run with `ANSIBLE_STDOUT_CALLBACK=debug` to surface any command output.  
- **Versioning** – Add a `requirements.yml` or a note about the minimum Ansible version (2.9+) and any OS‑level packages (`logrotate`, `gzip`, `rsync`).  
- **Future extensibility** – Suggest extracting the logrotate config into a Jinja template file (`templates/logrotate.conf.j2`) rather than embedding it in the `copy` task. This makes it easier to add comments or conditional sections later.

## 🧱 Mocks / Fakes
- The test suite correctly avoids touching real system logs by using `/tmp` directories, which is a good practice.  
- However, the `synchronize` module still performs a real rsync operation. In CI environments where network access is restricted, consider mocking it with the `ansible.builtin.command` module that copies files locally, or use the `delegate_to: localhost` pattern with `mode: pull` to keep everything on the same host.  
- The `ping` task is a lightweight way to verify reachability, but it will fail if the remote host does not have the `ping` module installed (e.g., minimal containers). You could replace it with a simple `wait_for` on port 22, which is more universally available.  

**Suggested mock for rsync in CI**
```yaml
- name: Mock rsync when running in CI
  when: lookup('env','CI') == 'true'
  block:
    - name: Copy rotated logs locally instead of rsync
      copy:
        src: "{{ item.path }}"
        dest: "{{ backup_destination }}/"
        mode: "0644"
      loop: "{{ rotated_logs.files }}"
```

---

Overall, the utility is well‑structured and the documentation is solid. Addressing the test reliability points, tightening the security handling of remote backups, and polishing the developer docs will make the contribution robust and easier for future maintainers to adopt.

polsala:
## 🤖 Review by GEMINI Agent

✅ What's solid
*   **Clear Structure and Documentation**: The `README.md` is comprehensive, detailing features, requirements, variables, and usage with a clear example. The variables table is particularly well-structured.
*   **Idempotency**: The playbook uses Ansible modules effectively to ensure idempotency where possible (e.g., `file` module for directory creation, `copy` for config deployment).
*   **Variable-driven Configuration**: Key parameters like `logrotate_path`, `log_files`, `rotate_days`, and backup settings are exposed as variables, making the playbook flexible and reusable.
*   **Isolated Testing**: The test playbook (`test_log_rotate.yml`) runs on `localhost` and uses temporary mock directories, ensuring it doesn't interfere with the actual system or require elevated privileges for testing.
*   **Dry-run for Logrotate**: The inclusion of a `logrotate -d` dry-run command is a good safety measure for debugging and understanding potential changes before applying them.

🧪 Tests
*   **Logrotate Naming Convention**: The test currently asserts for `app.log.1.gz`. While common, `logrotate` can use different naming schemes (e.g., `app.log-YYYYMMDD.gz` or `app.log.1` before compression).
    *   **Actionable Feedback**: Adjust the `find` pattern in `tests/test_log_rotate.yml` to be more resilient to `logrotate`'s naming variations, or ensure the `logrotate` configuration explicitly dictates the `app.log.1.gz` format. For example, change `patterns: "app.log.1.gz"` to `patterns: "app.log.*.gz"` in the `Verify that the mock log was rotated` task.

🔒 Security
*   **Backup User and Destination Credentials**: The `backup_user` and `backup_destination` variables are exposed directly in the playbook. For production use, these should be managed securely.
    *   **Actionable Feedback**: Add a note in the `README.md` or a comment in `src/log_rotate.yml` advising users to secure sensitive variables like `backup_destination` and `backup_user` credentials, possibly via Ansible Vault or environment variables.
*   **`postrotate` Script Review**: The `postrotate` script `/usr/lib/rsyslog/rsyslog-rotate` is executed with root privileges. While this is a standard system script, its integrity should be ensured. It's also specific to `rsyslog` and might not be universally applicable or the most robust way to signal a logging daemon.
    *   **Actionable Feedback**: Review the necessity and portability of the `/usr/lib/rsyslog/rsyslog-rotate` script. A more generic `systemctl reload rsyslog` or `killall -HUP rsyslogd` might be safer and more widely applicable, or consider if the `postrotate` action is even necessary if `logrotate` is managed by `cron`/`systemd` which handles daemon signaling.

🧩 Docs/DX
*   **`postrotate` Script Explanation**: The `postrotate` script `/usr/lib/rsyslog/rsyslog-rotate` is specific.
    *   **Actionable Feedback**: Add a note in the `README.md` or a comment in `src/log_rotate.yml` explaining the `postrotate` script's role and potential alternatives for different `rsyslog` configurations or other logging daemons.
*   **Logrotate Triggering Clarification**: The playbook explicitly runs `logrotate` twice (dry-run and actual run). While useful for immediate testing, the primary way `logrotate` is typically triggered is via a daily cron job or systemd timer. The current approach might lead to logs being rotated more frequently than intended if the playbook is run often.
    *   **Actionable Feedback**: Clarify in the `README.md` whether the playbook is intended to *deploy* the configuration and let the system's `cron`/`systemd` handle rotation, or to *force* an immediate rotation. If the latter, explain the implications. If the former, consider removing the explicit `command: logrotate` tasks from the main playbook, as deploying the config is sufficient.
*   **`find` task for `rotated_logs` Scope**: The `find` task for `rotated_logs` uses `paths: "/var/log"` and `age: "{{ rotate_days }}d"`. This might not perfectly align with the `log_files` variable if logs are in other locations, or with the exact logs just rotated. The `age` parameter might also pick up logs that are too old, not just the ones recently rotated.
    *   **Actionable Feedback**: Adjust the `find` task's `paths` to dynamically use the directories derived from the `log_files` variable. Re-evaluate the `age` parameter to ensure it correctly identifies the logs intended for backup (e.g., recently rotated logs, not just old ones). For example, consider looking for files modified within the last day after rotation, or simply all `.gz` files in the target directories.

🧱 Mocks/Fakes
*   **`ping` Task in Test Environment**: The `ping` task in the main playbook (`delegate_to: "{{ backup_destination.split(':')[0] }}"`) will attempt to ping `localhost` during the test. While this works for the current mock setup, if `backup_destination` were mocked to a non-existent remote host, the test would fail on the `ping` task, which might be an unnecessary dependency for a unit-style test.
    *   **Actionable Feedback**: Consider adding a `when: backup_enabled and backup_destination is search('^localhost:')` condition to the `ping` task in `src/log_rotate.yml` if the intent is to skip the `ping` for purely local mock backups during testing, or ensure `localhost` is always reachable in the test environment.

PR #4380, opened 2026-04-28
feat(gemini): A whimsical Bash script to identify and optionally clean up old, forgotten files, treating them as digital debris.

polsala:
## Implementation Summary
- **Utility**: nightly-digital-debris-duster
- **Provider**: gemini
- **Location**: `bash-utils/nightly-nightly-digital-debris-duste`
- **Files Created**: 3
- **Description**: A whimsical Bash script to identify and optionally clean up old, forgotten files, treating them as digital debris.

## Rationale

## Why safe to merge
- Utility is isolated to `bash-utils/nightly-nightly-digital-debris-duste`.

## Test Plan
- Follow the instructions in the generated README at `bash-utils/nightly-nightly-digital-debris-duste/README.md`
- Run tests located in `bash-utils/nightly-nightly-digital-debris-duste/tests/`

## Links

## Mock Justification

### DIFF
--- a/bash-utils/nightly-nightly-digital-debris-duste/README.md
+++ b/bash-utils/nightly-nightly-digital-debris-duste/README.md
@@ -1,0 +1,38 @@
+# Nightly Digital Debris Duster
+
+## Summary
+In the post-apocalyptic digital landscape, forgotten files accumulate like dust bunnies in a derelict server farm. The `nightly-digital-debris-duster` is a whimsical Bash script designed to help you identify and optionally scavenge (delete) these old, unused files, tidying up your digital wasteland.
+
+## Philosophy
+Every byte counts in the new world. This utility helps you reclaim precious storage by clearing out the digital detritus that no longer serves a purpose, making your systems leaner and more efficient for survival.
+
+## Usage
+
+To run the Digital Debris Duster, execute the script with optional arguments for the target directory, the age threshold, and the minimum size for debris.
+
+```bash
+bash src/debris_duster.sh [DIRECTORY] [DAYS_OLD] [MIN_SIZE_KB]
+```
+
+- `DIRECTORY`: (Optional) The path to the directory to scan. Defaults to the current directory (`.`).
+- `DAYS_OLD`: (Optional) Files older than this many days will be considered debris. Defaults to `30` days.
+- `MIN_SIZE_KB`: (Optional) Files smaller than this many kilobytes will be ignored. Defaults to `1` KB.
+
+### Examples
+
+1. **Scan the current directory for debris older than 30 days and larger than 1KB (default behavior):**
+   ```bash
+   bash src/debris_duster.sh
+   ```
+
+2. **Scan a specific directory (`/var/log/old_archives`) for files older than 90 days and larger than 100KB:**
+   ```bash
+   bash src/debris_duster.sh /var/log/old_archives 90 100
+   ```
+
+3. **Scan your home directory for any file older than 7 days, regardless of size (by setting MIN_SIZE_KB to 0):**
+   ```bash
+   bash src/debris_duster.sh ~/ 7 0
+   ```
+
+Upon identifying potential debris, the script will list the files and prompt you for confirmation before proceeding with deletion. Always review the list carefully before confirming!
--- a/bash-utils/nightly-nightly-digital-debris-duste/src/debris_duster.sh
+++ b/bash-utils/nightly-nightly-digital-debris-duste/src/debris_duster.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+
+# --- Configuration ---
+TARGET_DIR="${1:-.}"
+DAYS_OLD="${2:-30}"
+MIN_SIZE_KB="${3:-1}" # Minimum size in KB
+
+# --- Functions ---
+log_info() {
+    echo "✨ Integrator's Log: $*"
+}
+
+log_warning() {
+    echo "⚠️ Warning: $*" >&2
+}
+
+log_error() {
+    echo "❌ Error: $*" >&2
+    exit 1
+}
+
+confirm_action() {
+    local prompt="$1"
+    while true; do
+        echo -n "$prompt (y/n): "
+        read -r response # This 'read' can be mocked by the test script
+        case "$response" in
+            [yY][eE][sS]|[yY]) return 0 ;;
+            [nN][oO]|[nN]) return 1 ;;
+            *) echo "Invalid input. Please enter 'y' or 'n'." ;;
+        esac
+    done
+}
+
+# --- Main Logic ---
+main() {
+    if [[ ! -d "$TARGET_DIR" ]]; then
+        log_error "Target directory '$TARGET_DIR' does not exist."
+    fi
+
+    log_info "Scanning '$TARGET_DIR' for digital debris older than $DAYS_OLD days and larger than ${MIN_SIZE_KB}KB..."
+
+    # Find files:
+    # -type f: regular files
+    # -mtime +$DAYS_OLD: modified more than $DAYS_OLD days ago
+    # -size +${MIN_SIZE_KB}k: size greater than $MIN_SIZE_KB kilobytes
+    # -print0: null-terminated output for safety with filenames containing spaces/newlines
+    found_files=$(find "$TARGET_DIR" -type f -mtime +"$DAYS_OLD" -size "+${MIN_SIZE_KB}k" -print0)
+
+    if [[ -z "$found_files" ]]; then
+        log_info "No significant digital debris found. Your digital wasteland is surprisingly clean!"
+        return 0
+    fi
+
+    log_info "Identified the following digital debris:"
+    echo -e "$found_files" | xargs -0 -I {} echo "  - {}"
+
+    if confirm_action "Do you wish to scavenge (delete) this digital debris?"; then
+        log_info "Initiating debris scavenging..."
+        echo -e "$found_files" | xargs -0 rm -v # This 'rm' can be mocked by the test script
+        log_info "Digital debris scavenged. The wasteland is a bit tidier."
+    else
+        log_info "Digital debris left untouched. Perhaps it holds sentimental value."
+    fi
+}
+
+# Execute main function if script is not sourced
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
+    main "$@"
+fi
--- a/bash-utils/nightly-nightly-digital-debris-duste/tests/test_debris_duster.sh
+++ b/bash-utils/nightly-nightly-digital-debris-duste/tests/test_debris_duster.sh
@@ -1,0 +1,146 @@
+#!/bin/bash
+
+# Mock rationale: To ensure deterministic and offline testing,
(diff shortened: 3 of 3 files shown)


polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **Clear purpose & UX** – The script’s intent (find “digital debris” and optionally delete it) is well‑described in the README, and the interactive prompt makes the operation safe for users.  
- **Self‑contained utility** – All code lives under `bash-utils/nightly‑nightly‑digital‑debris‑duste/`; there are no external dependencies beyond standard Unix tools (`find`, `xargs`, `rm`).  
- **Good logging** – The `log_info`, `log_warning`, and `log_error` helpers give consistent, colour‑free output that can be piped or redirected.  
- **Test coverage intent** – The test suite tries to cover the main branches (no files, user declines, user accepts, too‑new, too‑small, invalid directory) and uses mocks to keep the tests deterministic and side‑effect‑free.  

## 🧪 Tests
| **Test harness sources the script** (`source ../src/debris_duster.sh "$@"`). When sourced, the `if [[ "${BASH_SOURCE[0]}" == "${0}" ]]` guard prevents `main` from running, so the tests never exercise the script logic. | Change the harness to invoke the script as a separate process, e.g. `bash ../src/debris_duster.sh "$@"`. This also isolates the mocked `rm`/`read` functions from the script’s own environment. |
| **Mocking `read`** – The script calls the builtin `read`. Overriding it with a function works, but the script uses `read -r response` which expects the builtin’s `-r` flag. The mock signature `read()` discards any arguments, which could hide future changes that rely on flags. | Implement the mock as `read() { REPLY="${MOCKED_READ_INPUT}"; }` **or** better, export a variable `MOCKED_READ_INPUT` and modify the script to call a wrapper function (`prompt_read`) that can be swapped in tests. |
| **`rm` mock** – The test overrides `rm` globally, which is fine because cleanup uses `/bin/rm`. However the mock does not preserve exit codes; if the script ever checks `$?` after `rm`, the mock would break that logic. | Return a success status explicitly: `rm() { echo "MOCKED_RM_CALL: $*" >> "$MOCKED_RM_LOG"; return 0; }`. |
| **`found_files` handling** – The script captures `find … -print0` output into a scalar variable, which strips the NUL delimiters. The subsequent `echo -e "$found_files" | xargs -0 …` works only for a single file without spaces. | Refactor to avoid command substitution for NUL‑separated data. Example: <br>```bash\nfind "$TARGET_DIR" -type f -mtime +"$DAYS_OLD" -size "+${MIN_SIZE_KB}k" -print0 |\n  while IFS= read -r -d '' file; do\n    echo "  - $file"\n    files+=("$file")\n  done\n```\nThen feed the array to `xargs -0 rm -v "${files[@]}"` or use `printf '%s\0' "${files[@]}" | xargs -0 rm -v`. This makes the script robust for many files and filenames containing spaces or newlines. |
| **Test reliability** – The tests create temporary files with `dd` and `touch`. If the CI runner has a very slow disk, the `dd` commands could become a bottleneck. | Use `truncate -s "${size_kb}K" "$path"` instead of `dd`; it’s faster and doesn’t depend on block‑device performance. |
| **Exit codes** – The script exits with `log_error` (which calls `exit 1`) on an invalid directory, but the test checks for the error message only. It would be useful to assert the exit status as well. | In the test harness capture `$?` after running the script and assert it equals `1` for error cases. |

## 🔒 Security
- **Argument sanitisation** – The script passes `$TARGET_DIR`, `$DAYS_OLD`, and `$MIN_SIZE_KB` directly to `find`. Since these are positional parameters, a malicious user could supply something like `"; rm -rf /"`; however `find` treats each argument as a separate token, so injection is unlikely. Still, adding a simple validation step improves robustness: <br>```bash\n[[ "$DAYS_OLD" =~ ^[0-9]+$ ]] || log_error "DAYS_OLD must be a non‑negative integer."\n[[ "$MIN_SIZE_KB" =~ ^[0-9]+$ ]] || log_error "MIN_SIZE_KB must be a non‑negative integer."\n``` |
- **Deletion safety** – The script prompts before deletion, which is good. To further protect against accidental mass deletion, consider adding a `--dry-run` flag that only lists files without prompting. |
- **Path traversal** – The script operates on the directory supplied by the user. If the user passes `..` or a symlink that points outside an intended sandbox, the script will still act on those paths. If the utility is ever used in a constrained environment, you may want to resolve the absolute path and optionally enforce a whitelist. |
- **Use of `rm -v`** – No `-f` flag is used, which is safe. However, if the script is ever extended to accept additional `rm` options, ensure they are whitelisted. |

- **Folder name typo** – The directory is named `nightly‑nightly‑digital‑debris‑duste` (missing the trailing “r”). This typo propagates to the PR title and README links. Rename the folder to `nightly-digital-debris-duster` and update references. |
- **README command examples** – The examples use `bash src/debris_duster.sh …`. Since the script is already executable (`#!/usr/bin/env bash`), you can run it directly: `./src/debris_duster.sh …`. Mention both ways for clarity. |
- **Parameter documentation** – The README lists defaults but does not state the accepted range (e.g., `DAYS_OLD` must be a positive integer). Add a “Constraints” subsection. |
- **Help flag** – Consider adding a `-h/--help` option that prints usage information. This improves discoverability for new users. |
- **Shebang portability** – Use `#!/usr/bin/env bash` instead of `#!/bin/bash` to work on systems where Bash is not in `/bin`. |
- **License header** – If the repository requires a license header on new scripts, add one at the top of `debris_duster.sh`. |

## 🧱 Mocks / Fakes
- **Scope of mocks** – Overriding `rm` and `read` globally works for the current test suite, but any future test that needs the real `rm` (e.g., cleanup) must remember to call `/bin/rm`. This can be error‑prone. |
  - **Recommendation**: Export a variable that tells the script to use a mock deletion function, e.g., `DEBRIS_DUSTER_MOCK_RM=1`. Inside the script, replace `rm -v` with a wrapper: <br>```bash\nif [[ -n "$DEBRIS_DUSTER_MOCK_RM" ]]; then\n  mock_rm -v \"$file\"\nelse\n  rm -v \"$file\"\nfi\n```<br>Then the test can set the env var and provide its own `

polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **Clear purpose & UX** – The script’s intent (find “digital debris” and optionally delete it) is well‑articulated in the README, and the interactive prompt makes the operation safe for users.  
- **Self‑contained utility** – All code lives under a single folder, with no external dependencies, making it easy to drop into other projects.  
- **Comprehensive test suite** – Six test cases cover the main branches (no debris, user declines, user accepts, too‑new, too‑small, invalid directory).  
- **Good logging style** – The `log_*` helpers give consistent, colour‑free output that can be piped or redirected without breaking scripts.  

## 🧪 Tests
| Tests mock `rm` and `read` by defining shell functions, but they are **not exported** (`export -f rm read`). Consequently `xargs … rm` still invokes the real `/bin/rm`, potentially deleting files on the host. | Export the mocks so child processes see them, e.g.:<br>`export -f rm read`<br>or use a wrapper script (`mock_rm.sh`) and point `xargs` to it with `-I {} ./mock_rm.sh {}`. |
| The mocked `read` function ignores the `-r response` arguments and only sets `REPLY`. The script expects the variable `response` to be populated, so the confirmation loop could become infinite or always treat the input as invalid. | Implement the mock to honour the variable name:<br>```bash\nread() {\n  local var_name=\"$1\"\n  printf -v \"$var_name\" \"%s\" \"$MOCKED_READ_INPUT\"\n}\n``` |
| `found_files=$(find … -print0)` stores a NUL‑separated list in a plain string. Later `echo -e "$found_files"` loses the NULs, breaking filenames with spaces or newlines. | Use an array or process substitution instead of a variable, e.g.:<br>```bash\nmapfile -d '' -t files < <(find \"$TARGET_DIR\" -type f -mtime +\"$DAYS_OLD\" -size +\"${MIN_SIZE_KB}k\" -print0)\nif (( ${#files[@]} == 0 )); then …\nfi\nprintf '%s\n' \"${files[@]}\"\n``` |
| Tests rely on `dd` to create files of a specific size. On some CI runners `dd` may be unavailable or restricted. | Consider using `truncate -s "${size_kb}K" "$path"` which is faster and more portable. |
| Test cleanup uses the real `/bin/rm` after the suite finishes. If a previous test accidentally invoked the real `rm`, leftover files could remain. | Perform cleanup **before** each test (or in a `trap`), and always use the real binary via an absolute path (`/bin/rm -rf "$TEST_DIR"`). |

## 🔒 Security
- **Path handling** – The script trusts the user‑supplied `TARGET_DIR`. If a malicious caller passes something like `"/tmp/../etc"` the `find` command will still resolve it correctly, but it’s good practice to canonicalise the path (`realpath`) before use to avoid accidental traversal.  
- **Command injection** – Arguments are passed directly to `find` and `xargs`. Since they are not interpolated into a shell string, the risk is low, but adding `set -u` and quoting all variable expansions (`"$TARGET_DIR"`) already mitigates most issues.  
- **Exit on error** – The script currently exits only on a missing directory (`log_error`). Adding `set -euo pipefail` at the top will abort on unexpected failures (e.g., permission errors while reading files).  
- **Deletion safety** – Even with the confirmation prompt, a scripted run could auto‑answer “y”. Consider adding a `--dry-run` flag that lists files without invoking `rm`, or require an explicit `--force` flag for non‑interactive usage.  

- **Folder name typo** – The directory is `nightly-nightly-digital-debris-duste` (missing the trailing “r”). Align the folder name with the utility (`nightly-digital-debris-duster`) to avoid confusion.  
- **README usage path** – The examples use `bash src/debris_duster.sh`. When the README is viewed from the repository root, that path is incorrect; the script lives under `bash-utils/nightly‑.../src`. Update the examples to either: <br>`bash bash-utils/nightly‑.../src/debris_duster.sh …` <br>or instruct users to `cd` into the utility folder first.  
- **Help output** – Adding a `--help` flag (or handling `-h/--help`) that prints usage and defaults would improve discoverability, especially for non‑interactive environments.  
- **Versioning** – Include a `VERSION` variable or comment header (e.g., `# v1.0.0`) so downstream scripts can check compatibility.  
- **Shellcheck compliance** – Run `shellcheck` on the script; it will flag issues such as unquoted variables, missing `local` declarations, and the use of `echo -e` for raw data. Addressing these warnings will make the script more robust.  

## 🧱 Mocks / Fakes
- **Mock export** – As noted in the Tests section, export the mock functions so they survive into subshells spawned by `xargs`.  
- **Isolation of external commands** – Instead of globally overriding `rm`, consider injecting the command name via a variable (e.g., `RM_CMD=${RM_CMD:-rm}`) and using `$RM_CMD` inside the script. Tests can then set `RM_CMD=mock_rm` without needing function export.  
- **Read mock robustness** – The current mock replaces the builtin `read` entirely. A less invasive approach is to wrap the prompt in a function (`prompt_user`) that can be overridden in tests, keeping the builtin `read` untouched for other scripts.  

---

**Overall impression:** The utility is a fun and useful addition, and the test coverage is commendable. Addressing the mock/export issues, tightening the handling of NUL‑separated file lists, and polishing the documentation will make the contribution production‑ready and safer for downstream users.

PR #4462, opened 2026-05-06
feat(gemini): A Go-based concurrent network probe for checking service availability and latency.

polsala:
## Implementation Summary
- **Utility**: nightly-go-network-probe
- **Provider**: gemini
- **Location**: `go-utils/nightly-nightly-go-network-probe`
- **Files Created**: 3
- **Description**: A Go-based concurrent network probe for checking service availability and latency.

## Rationale

## Why safe to merge
- Utility is isolated to `go-utils/nightly-nightly-go-network-probe`.

## Test Plan
- Follow the instructions in the generated README at `go-utils/nightly-nightly-go-network-probe/README.md`
- Run tests located in `go-utils/nightly-nightly-go-network-probe/tests/`

## Links

## Mock Justification

### DIFF
--- a/go-utils/nightly-nightly-go-network-probe/README.md
+++ b/go-utils/nightly-nightly-go-network-probe/README.md
@@ -1,0 +1,30 @@
+# nightly-go-network-probe
+
+A whimsical yet useful Go utility for concurrently probing network services. It checks for service availability and measures latency, providing a quick overview of your network's health.
+
+## Philosophy
+
+Inspired by the need for a simple, fast, and concurrent way to check if services are up and responsive, especially in a post-apocalyptic scenario where reliable communication is key. Built with Go for its excellent concurrency primitives and performance.
+
+## Usage
+
+Compile the Go program and run it with a list of target hosts and ports.
+
+```bash
+# Example:
+./nightly-go-network-probe google.com:80 example.com:443 1.1.1.1:53
+```
+
+The output will show the status (UP/DOWN) and latency for each probed service.
+
+## How it Works
+
+The utility takes host:port combinations as command-line arguments. For each target, it launches a goroutine to perform a TCP connection attempt. It measures the time taken for the connection to establish or fail, reporting the result and latency.
+
+## Testing
+
+Tests are included to ensure the functionality of the probe. They use mocked network responses to provide deterministic and offline testing.
+
+## Contributing
+
+Feel free to fork this repository and submit pull requests with improvements or new features. All contributions are welcome!
--- a/go-utils/nightly-nightly-go-network-probe/src/main.go
+++ b/go-utils/nightly-nightly-go-network-probe/src/main.go
@@ -1,0 +1,71 @@
+package main
+
+import (
+	"fmt"
+	"net"
+	"os"
+	"sync"
+	"time"
+)
+
+// ProbeResult holds the outcome of a network probe.
+type ProbeResult struct {
+	Target  string
+	Status  string
+	Latency time.Duration
+	Error   error
+}
+
+// probeTarget attempts to connect to a given host:port and measures latency.
+func probeTarget(target string, wg *sync.WaitGroup, results chan<- ProbeResult) {
+	defer wg.Done()
+
+	start := time.Now()
+	conn, err := net.DialTimeout("tcp", target, 5*time.Second) // 5-second timeout
+
+	latency := time.Since(start)
+
+	result := ProbeResult{Target: target}
+
+	if err != nil {
+		result.Status = "DOWN"
+		result.Error = err
+	} else {
+		result.Status = "UP"
+		result.Latency = latency
+		conn.Close() // Close the connection immediately after successful probe
+	}
+
+	results <- result
+}
+
+func main() {
+	if len(os.Args) < 2 {
+		fmt.Println("Usage: nightly-go-network-probe <host:port> [<host:port> ...]")
+		os.Exit(1)
+	}
+
+	targets := os.Args[1:]
+	var wg sync.WaitGroup
+	results := make(chan ProbeResult, len(targets))
+
+	fmt.Println("Starting network probes...")
+
+	for _, target := range targets {
+		wg.Add(1)
+		go probeTarget(target, &wg, results)
+	}
+
+	wg.Wait()
+	close(results)
+
+	fmt.Println("\n--- Probe Results ---")
+	for result := range results {
+		if result.Status == "UP" {
+			fmt.Printf("%s: %s (Latency: %s)\n", result.Target, result.Status, result.Latency.Round(time.Millisecond))
+		} else {
+			fmt.Printf("%s: %s (Error: %v)\n", result.Target, result.Status, result.Error)
+		}
+	}
+	fmt.Println("---------------------")
+}
--- a/go-utils/nightly-nightly-go-network-probe/tests/test_main.go
+++ b/go-utils/nightly-nightly-go-network-probe/tests/test_main.go
@@ -1,0 +1,136 @@
+package main
+
+import (
+	"net"
+	"testing"
+	"time"
+)
+
+// MockDialer is a mock implementation of net.Dialer.
+type MockDialer struct {
+	DialFunc func(network, address string) (net.Conn, error)
+}
+
+func (m *MockDialer) Dial(network, address string) (net.Conn, error) {
+	if m.DialFunc != nil {
+		return m.DialFunc(network, address)
+	}
+	return nil, nil
+}
+
+// MockConn is a mock implementation of net.Conn.
+type MockConn struct {}
+
+func (m *MockConn) Read(b []byte) (n int, err error) { return 0, nil }
+func (m *MockConn) Write(b []byte) (n int, err error) { return 0, nil }
+func (m *MockConn) Close() error { return nil }
+func (m *MockConn) LocalAddr() net.Addr { return nil }
+func (m *MockConn) RemoteAddr() net.Addr { return nil }
+func (m *MockConn) SetDeadline(t time.Time) error { return nil }
+func (m *MockConn) SetReadDeadline(t time.Time) error { return nil }
+func (m *MockConn) SetWriteDeadline(t time.Time) error { return nil }
+
+// Mock rationale: Replace the global net.Dialer with our mock for deterministic testing.
+// This allows us to control the outcome of network connections without actual network calls.
+var originalDialer = net.Dial
+
+func setupMockDialer(dialFunc func(network, address string) (net.Conn, error)) {
+	net.Dial = func(network, address string) (net.Conn, error) {
+		return (&MockDialer{DialFunc: dialFunc}).Dial(network, address)
+	}
+}
+
+func restoreDialer() {
+	net.Dial = originalDialer
+}
+
+func TestProbeTarget_Success(t *testing.T) {
+	defer restoreDialer()
+
+	// Mock rationale: Simulate a successful TCP connection.
+	setupMockDialer(func(network, address string) (net.Conn, error) {
+		// Simulate a small delay to mimic latency.
+		time.Sleep(10 * time.Millisecond)
+		return &MockConn{}, nil
+	})
+
+	var wg sync.WaitGroup
+	results := make(chan ProbeResult, 1)
+
+	wg.Add(1)
+	go probeTarget("example.com:80", &wg, results)
+	wg.Wait()
+	close(results)
+
+	result := <-results
+
+	if result.Status != "UP" {
+		t.Errorf("Expected status UP, got %s", result.Status)
+	}
+
+	if result.Error != nil {
+		t.Errorf("Expected no error, got %v", result.Error)
+	}
+
+	if result.Latency < 0 {
+		t.Errorf("Expected positive latency, got %s", result.Latency)
+	}
+}
+
+func TestProbeTarget_Failure(t *testing.T) {
+	defer restoreDialer()
+
+	// Mock rationale: Simulate a failed TCP connection.
+	setupMockDialer(func(network, address string) (net.Conn, error) {
+		return nil, &net.OpError{Op: "dial", Net: "tcp", Addr: nil, Err: fmt.Errorf("connection refused")}
+	})
+
+	var wg sync.WaitGroup
+	results := make(chan ProbeResult, 1)
+
+	wg.Add(1)
+	go probeTarget("nonexistent.com:8080", &wg, results)
+	wg.Wait()
+	close(results)
+
+	result := <-results
+
+	if result.Status != "DOWN" {
+		t.Errorf("Expected status DOWN, got %s", result.Status)
+	}
+
+	if result.Error == nil {
+		t.Errorf("Expected an error, got nil")
+	}
+}
+
+func TestProbeTarget_Timeout(t *testing.T) {
+	// Mock rationale: Simulate a connection timeout.
+	// We don't need to restore the dialer here because the main function's timeout
+	// will handle this scenario, and our mock will just return an error.
+	setupMockDialer(func(network, address string) (net.Conn, error) {
+		// Simulate a delay longer than the DialTimeout in probeTarget.
+		time.Sleep(6 * time.Second)
+		return nil, &net.OpError{Op: "dial", Net: "tcp", Addr: nil, Err: fmt.Errorf("i/o timeout")}
+	})
+
+	var wg sync.WaitGroup
+	results := make(chan ProbeResult, 1)
+
+	wg.Add(1)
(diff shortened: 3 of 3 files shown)


polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **Concurrency model** – The utility spins up a goroutine per target and uses a `sync.WaitGroup` plus a buffered channel to collect results. This pattern is simple, dead‑lock‑free and scales well with the number of probes.  
- **Clear output** – The result printing distinguishes *UP* and *DOWN* cases and formats latency with millisecond precision, which is exactly what a quick health‑check tool needs.  
- **Self‑contained** – All source files live under `go-utils/nightly-nightly-go-network-probe`; there are no external dependencies beyond the Go standard library, making the utility easy to vendor or copy into other repos.  
- **README tone** – The README gives a concise description, a one‑liner usage example, and a short “How it Works” section that helps a newcomer understand the probe’s internals.

## 🧪 Tests
The test suite aims to provide deterministic, offline verification, but it currently cannot compile or run as‑is.

| Issue | Why it matters | Suggested fix |
|-------|----------------|---------------|
| **Missing imports** – `sync` and `fmt` are used in the tests but not imported. | The test file fails to compile, so no coverage is achieved. | Add `import ("fmt"; "sync")` alongside the existing imports. |
| **Attempted reassignment of `net.Dial`** – `net.Dial` is a *function*, not a variable, so `net.Dial = …` is illegal. | The mock never takes effect and the code won’t build. | Introduce a package‑level variable that holds the dialing function, e.g.:<br>```go\n// in main.go\nvar dialer = net.DialTimeout\n\nfunc probeTarget(target string, wg *sync.WaitGroup, results chan<- ProbeResult) {\n    conn, err := dialer(\"tcp\", target, 5*time.Second)\n    // …\n}\n```<br>Then in tests you can replace `dialer` with a mock implementation. |
| **Mocking the wrong function** – `probeTarget` calls `net.DialTimeout`, but the test replaces `net.Dial`. | Even if reassignment were possible, the probe would still use the real timeout function, making the mock ineffective. | After introducing the `dialer` variable (see above), replace it in tests: `original := dialer; dialer = mockFunc; defer func(){dialer = original}()`. |
| **Latency measurement on failure** – The current implementation only records latency on success. | Tests that simulate a timeout or immediate failure cannot assert on latency, limiting coverage. | Record `latency := time.Since(start)` before the error check and always set `result.Latency = latency`. This also gives useful timing info for failures. |
| **Test naming & coverage** – The three tests cover success, failure, and timeout, which is good, but they could be more expressive. | Clearer intent helps future contributors understand edge cases. | Rename to `TestProbeTarget_Succeeds`, `TestProbeTarget_ReturnsErrorOnRefused`, `TestProbeTarget_HandlesTimeout`. Add a table‑driven test to iterate over multiple mock scenarios. |
| **Deterministic latency** – The “success” mock sleeps 10 ms, which can be flaky on CI runners. | Flaky tests erode confidence. | Instead of `time.Sleep`, return a mock connection immediately and set a fixed latency value in the mock (e.g., `result.Latency = 10 * time.Millisecond`). Or use a deterministic clock via an interface. |

### Minimal test‑ready refactor example
```go
// main.go (excerpt)
var dialer = net.DialTimeout // replaceable in tests

func probeTarget(target string, wg *sync.WaitGroup, results chan<- ProbeResult) {
    defer wg.Done()
    start := time.Now()
    conn, err := dialer("tcp", target, 5*time.Second)
    latency := time.Since(start)

    r := ProbeResult{Target: target, Latency: latency}
    if err != nil {
        r.Status = "DOWN"
        r.Error = err
    } else {
        r.Status = "UP"
        conn.Close()
    }
    results <- r
}
```

```go
// test_main.go (excerpt)
func TestProbeTarget_Success(t *testing.T) {
    orig := dialer
    dialer = func(_, _ string, _ time.Duration) (net.Conn, error) {
        return &MockConn{}, nil // no sleep → deterministic latency
    }
    defer func() { dialer = orig }()

    // … rest of test unchanged …
}
```

## 🔒 Security
- **Input validation** – The program accepts arbitrary `host:port` strings from the command line and passes them directly to `net.DialTimeout`. While the standard library sanitises the address format, a malicious user could supply extremely long strings or malformed inputs that trigger panics or excessive memory allocation.  
  *Action*: Validate each argument with `net.SplitHostPort` and enforce reasonable length limits (e.g., host ≤ 255 bytes, port within 1‑65535). Return a friendly error instead of panicking.  

- **Rate limiting / abuse prevention** – As a network scanner, the utility could be mis‑used to flood internal services. Adding a simple per‑second limit (e.g., via a token bucket) or a `-max-concurrency` flag would give operators control over the probe’s aggressiveness.  

- **TLS / encrypted probes** – The current implementation only supports raw TCP. If the utility is later extended to probe HTTPS or other TLS services, ensure proper certificate verification or provide an explicit `-insecure` flag with a clear warning.

- **Build instructions** – The README jumps straight to running the compiled binary. Adding a “Build” section (`go build -o nightly-go-network-probe ./src`) helps newcomers who may not know the module layout.  
- **Command‑line flags** – Using the `flag` package (or a lightweight CLI library) would let users customize timeout, concurrency, and output format without editing source. Document these flags in the README.  
- **Example output** – Show a sample run with both *UP* and *DOWN* entries, including the latency format. This makes the tool’s output expectations explicit.  
- **Go module file** – The repository currently lacks a `go.mod`. Adding one (e.g., `module github.com/yourorg/go-utils/nightly-go-network-probe`) clarifies the module path and Go version, and it enables reproducible builds.  
- **License header** – Include a short SPDX identifier at the top of each source file (e.g., `// SPDX-License-Identifier: Apache-2.0`) to avoid ambiguity about reuse rights.

## 🧱 Mocks / Fakes
- The test suite introduces `MockDialer` and `MockConn`, which is a good start, but the current design forces the mock to replace a non‑replaceable function. Refactoring the probe to accept a dial function (as shown above) decouples the production code from the network stack and makes the mocks truly effective.  

- **Avoid global state** – Instead of a package‑level `dialer` variable, consider passing the dial function as an argument to `probeTarget` (or to a higher‑level orchestrator). This keeps the function pure and eliminates the need for restore/cleanup logic in every test.  

- **Reuse Go’s `nettest` utilities** – The standard library provides `nettest.NewLocalListener` for creating in‑process listeners. For integration‑style tests you could spin up a temporary TCP server that accepts connections, then verify that the probe reports *UP* with realistic latency. This reduces the amount of hand‑rolled mock code.

---

**Overall impression:** The utility’s core idea and concurrency implementation are solid, and the README gives a friendly introduction. The primary blocker is the test suite, which currently cannot compile and does not correctly mock the network layer. Refactoring the dialing mechanism to be injectable will unlock reliable, deterministic tests and improve the code’s testability. Adding a few defensive checks, build instructions, and optional CLI flags will round out the developer experience and make the tool safer for broader use.

polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **Concurrency model** – The probe spins up a goroutine per target, uses a `sync.WaitGroup` and a buffered channel to collect results. This pattern is dead‑lock‑free, easy to read, and scales with the number of probes.  
- **Result handling** – The `ProbeResult` struct cleanly separates status, latency, and error information, making the output formatting straightforward.  
- **CLI ergonomics** – The program validates that at least one `host:port` argument is supplied and prints a concise usage line when called incorrectly.  
- **Timeout handling** – `net.DialTimeout` with a 5 s deadline prevents a single slow endpoint from stalling the whole run.  

## 🧪 Tests
The test suite aims to verify success, failure, and timeout paths, but it currently cannot compile or run as‑is.

| Issue | Why it matters | Suggested fix |
|-------|----------------|---------------|
| **Monkey‑patching `net.Dial`** – `net.Dial` (and `net.DialTimeout`) are regular functions, not assignable variables. The code `net.Dial = …` will not compile. | Tests never intercept the real network call, so they either hit the network or fail to build. | Refactor `probeTarget` to accept a *dial function* (e.g., `type dialFn func(network, address string) (net.Conn, error)`) and inject it from the test. Example: <br>```go\nfunc probeTarget(target string, wg *sync.WaitGroup, results chan<- ProbeResult, dial dialFn) { … conn, err := dial(\"tcp\", target) }\n```<br>In production call `probeTarget(t, &wg, results, net.DialTimeout)`. |
| **Missing imports** – `sync`, `fmt`, and `time` are used in the test file but not imported. | Compilation error. | Add the required imports at the top of `test_main.go`. |
| **File naming** – The test file is named `test_main.go`. Go only runs files that match `*_test.go`. | Tests will be ignored by `go test`. | Rename to `main_test.go` (or any `*_test.go`). |
| **Long sleeps in timeout test** – `time.Sleep(6 * time.Second)` makes the test slow and flaky. | CI pipelines become unnecessarily long; test reliability suffers. | Simulate a timeout by returning a `net.OpError` with `Timeout()` true, without sleeping. Example: <br>```go\nreturn nil, &net.OpError{Op: \"dial\", Net: \"tcp\", Err: &os.SyscallError{Syscall: \"connect\", Err: syscall.ETIMEDOUT}}\n``` |
| **Error construction uses `fmt` without import** – `fmt.Errorf` appears in the failure mock. | Compile error. | Import `fmt` or use `errors.New`. |
| **`originalDialer` captures `net.Dial`** – Even if reassignment were possible, restoring the original function is unsafe for parallel test runs. | Could leak state between tests. | With the injected dial function approach, no global state mutation is needed. |
| **Test coverage** – Only three scenarios are covered, and they are not table‑driven. | Harder to extend and maintain. | Convert to a table‑driven test that iterates over a slice of structs containing `name`, `mockFn`, `expectedStatus`, etc. This reduces duplication and makes adding new cases trivial. |

### Minimal example of a refactored test

```go
type dialFn func(network, address string) (net.Conn, error)

func TestProbeTarget(t *testing.T) {
    cases := []struct {
        name          string
        mockDial      dialFn
        wantStatus    string
        wantErrSubstr string
    }{
        {
            name: "successful connection",
            mockDial: func(_, _ string) (net.Conn, error) {
                return &MockConn{}, nil
            },
            wantStatus: "UP",
        },
        {
            name: "connection refused",
            mockDial: func(_, _ string) (net.Conn, error) {
                return nil, &net.OpError{Op: "dial", Net: "tcp", Err: fmt.Errorf("connection refused")}
            },
            wantStatus:    "DOWN",
            wantErrSubstr: "connection refused",
        },
        {
            name: "timeout",
            mockDial: func(_, _ string) (net.Conn, error) {
                return nil, &net.OpError{Op: "dial", Net: "tcp", Err: fmt.Errorf("i/o timeout")}
            },
            wantStatus:    "DOWN",
            wantErrSubstr: "i/o timeout",
        },
    }

    for _, tc := range cases {
        t.Run(tc.name, func(t *testing.T) {
            var wg sync.WaitGroup
            results := make(chan ProbeResult, 1)

            wg.Add(1)
            go probeTarget("example.com:80", &wg, results, tc.mockDial)
            wg.Wait()
            close(results)

            res := <-results
            if res.Status != tc.wantStatus {
                t.Fatalf("expected %s, got %s", tc.wantStatus, res.Status)
            }
            if tc.wantErrSubstr != "" && !strings.Contains(res.Error.Error(), tc.wantErrSubstr) {
                t.Fatalf("expected error containing %q, got %v", tc.wantErrSubstr, res.Error)
            }
        })
    }
}
```

## 🔒 Security
- **Input validation** – The program trusts any string passed on the command line. A malformed `host:port` (e.g., `":/../etc/passwd"`) will cause `net.DialTimeout` to return an error, but it’s better to validate the format early and provide a clear error message. Consider using `net.SplitHostPort` to verify the argument structure.  
- **DNS rebinding / spoofing** – Since the utility performs raw TCP connections, an attacker could supply a hostname that resolves to an internal IP. If this tool is ever run with elevated privileges, add a flag to restrict probing to a whitelist or to enforce IPv4/IPv6 only.  
- **Resource exhaustion** – No hard limit on the number of concurrent goroutines; a malicious user could invoke the binary with thousands of targets, exhausting memory. Add a configurable concurrency ceiling (e.g., a semaphore or worker pool).  

- **Build instructions** – The README mentions “Compile the Go program” but does not show the exact command (`go build -o nightly-go-network-probe ./src`). Add a short “Build” section.  
- **Exit codes** – The binary always exits with `0` (except for missing arguments). It would be useful to return a non‑zero code when any probe fails, enabling CI usage (`if any DOWN → exit 1`). Document this behavior.  
- **Flag support** – Hard‑coding the timeout (5 s) limits flexibility. Expose a `-timeout` flag (e.g., `time.Duration` parsed via `flag.Duration`). Document the flag in the README.  
- **Example output** – Include a sample run showing both UP and DOWN cases, plus the exit code, to set expectations for users.  
- **License** – The repository currently lacks a license file; adding an MIT/Apache‑2.0 header clarifies reuse rights.  

## 🧱 Mocks / Fakes
- The current test suite attempts to mock the network layer but does so by reassigning `net.Dial`, which is not possible. Refactoring to inject a dial function (as described in the **Tests** section) eliminates the need for global monkey‑patching and makes the code naturally testable.  
- The `MockConn` implementation satisfies `net.Conn` but returns `nil` for `LocalAddr`/`RemoteAddr`. That’s fine for the current tests, but if future tests need address information, consider returning stub `net.Addr` values.  
- No external mocking frameworks are required; simple function injection keeps the dependency surface minimal.  

---

### TL;DR Action items
1. **Refactor `probeTarget`** to accept a dial function (or a `net.Dialer`) so tests can inject deterministic behavior.  
2. **Fix the test file**: rename to `*_test.go`, add missing imports, and replace the long `Sleep` with immediate error returns.  
3. **Add input validation** (use `net.SplitHostPort`) and a configurable concurrency limit.  
4. **Expose a `-timeout` flag** and document build steps, exit codes, and sample output in the README.  
5. **Add a license** and consider returning a non‑zero exit code when any probe reports DOWN.  

Implementing these changes will make the utility production‑ready, improve test reliability, and provide a smoother developer experience for future contributors.

polsalaTipoDeIncognito (approved):

polsala:

✅ **What's solid**

- **Concurrency model**: The utility effectively uses goroutines per target with `sync.WaitGroup` and a buffered channel for collecting results. This is a clean, scalable pattern that avoids deadlocks.
- **Clear output formatting**: Results distinguish *UP* and *DOWN* states clearly and present latency in a readable millisecond format.
- **Self-contained scope**: The utility resides in its own directory and does not interfere with existing code, making it low-risk to merge.
- **Accompanying tests and docs**: Includes both unit tests and a README, which is commendable for generated utilities.

🧪 **Tests**

- **Test coverage is present but basic**:
  - Tests cover success, failure, and timeout cases, which is good.
  - However, the tests are tightly coupled to global state via `net.Dial` replacement, which can lead to flakiness or side effects if not handled carefully.
- **Mocking strategy has risks**:
  - Replacing `net.Dial` globally can cause issues if tests run in parallel or if other parts of the codebase rely on `net.Dial`.
  - Consider using dependency injection instead of global replacement. For example:
    ```go
    // Define an interface for dialing
    type Dialer interface {
        Dial(network, address string) (net.Conn, error)
    }

    // Use it in probeTarget
    func probeTarget(dialer Dialer, target string, wg *sync.WaitGroup, results chan<- ProbeResult) {
        ...
        conn, err := dialer.Dial("tcp", target)
        ...
    }
    ```
  - This allows passing in a mock dialer during testing without modifying global state.

🔒 **Security**

- No sensitive data or credentials are introduced—✅ safe in that regard.
- The utility performs only TCP connection attempts, which is a safe operation when used in a controlled context.
- **Potential DoS vector**:
  - If invoked with a large number of targets, it could open many goroutines and connections. Consider documenting or enforcing limits on the number of concurrent probes.

🧩 **Docs/DX (Developer Experience)**

- **README is helpful but could be more precise**:
  - The usage example assumes the binary is named `./nightly-go-network-probe`, but the actual file is `main.go`. Clarify that users should `go run src/main.go` or build it first.
  - Include example output in the README to improve clarity.
- **Missing build instructions**:
  - Add a section like:
    ```bash
    go build -o nightly-go-network-probe src/main.go
    ```
- **Consider adding flags for configurability**:
  - Timeout duration, output format (JSON?), or concurrency limits would improve usability.

🧱 **Mocks/Fakes**

- **Global state manipulation in tests is a red flag**:
  - As noted above, replacing `net.Dial` globally can cause side effects.
  - The current mock strategy works for now but is not a best practice for larger projects.
- **MockConn implementation is minimal but sufficient**:
  - It satisfies the `net.Conn` interface, which is good.
  - Could be extended with behavior simulation (e.g., delayed reads/writes) if needed in future tests.

---

### Summary

This is a well-structured and useful utility with a solid foundation. The concurrency model is clean, and the inclusion of tests and documentation is a big plus. However, the testing strategy should be refactored to avoid global state manipulation. Improving the README with clearer build/run instructions and example output would also enhance usability.

PR #5010, opened 2026-06-26
feat(gemini): A containerized utility to sweep a mounted directory for digital cobwebs like temporary files, large artifacts, and sensitive patterns.

polsala:
## Implementation Summary
- **Utility**: nightly-cobweb-sweeper
- **Provider**: gemini
- **Location**: `docker-tools/nightly-nightly-cobweb-sweeper`
- **Files Created**: 4
- **Description**: A containerized utility to sweep a mounted directory for digital cobwebs like temporary files, large artifacts, and sensitive patterns.

## Rationale

## Why safe to merge
- Utility is isolated to `docker-tools/nightly-nightly-cobweb-sweeper`.

## Test Plan
- Follow the instructions in the generated README at `docker-tools/nightly-nightly-cobweb-sweeper/README.md`
- Run tests located in `docker-tools/nightly-nightly-cobweb-sweeper/tests/`

## Links

## Mock Justification

### DIFF
--- a/docker-tools/nightly-nightly-cobweb-sweeper/Dockerfile
+++ b/docker-tools/nightly-nightly-cobweb-sweeper/Dockerfile
@@ -1,0 +1,18 @@
+# Use a minimal base image with common utilities
+FROM alpine:3.18
+
+# Install necessary tools: bash, findutils, grep, coreutils (for du)
+RUN apk add --no-cache bash findutils grep coreutils
+
+# Copy the scanner script into the container
+COPY src/cobweb_scanner.sh /usr/local/bin/cobweb_scanner.sh
+
+# Make the script executable
+RUN chmod +x /usr/local/bin/cobweb_scanner.sh
+
+# Set the entrypoint to the scanner script
+# This allows passing arguments directly to the script, e.g., the target directory
+ENTRYPOINT ["/usr/local/bin/cobweb_scanner.sh"]
+
+# Default command if no arguments are provided (scans /scan_target)
+CMD ["/scan_target"]
--- a/docker-tools/nightly-nightly-cobweb-sweeper/README.md
+++ b/docker-tools/nightly-nightly-cobweb-sweeper/README.md
@@ -1,0 +1,43 @@
+# Nightly Cobweb Sweeper
+
+A containerized utility designed to help you keep your project directories tidy by identifying "digital cobwebs" – temporary files, large build artifacts, empty directories, and potentially sensitive configuration files. Run it against any directory to get a quick report on potential cleanup opportunities.
+
+## Features
+
+*   **Temporary File Detection**: Flags files ending with common temporary extensions (`.tmp`, `.bak`, `~`, `#*#`).
+*   **Large File Identification**: Reports files exceeding a configurable size threshold (default: 10MB).
+*   **Empty Structure Detection**: Finds empty files and directories.
+*   **Sensitive Pattern Scan**: Identifies common sensitive file names like `.env` or `id_rsa`.
+*   **Containerized**: Runs in an isolated Docker container, ensuring consistent scanning environment and no host dependencies other than Docker itself.
+
+## Usage
+
+1.  **Build the Docker Image**:
+    ```bash
+    docker build -t nightly-cobweb-sweeper .
+    ```
+
+2.  **Scan a Directory**:
+    Navigate to the directory you want to scan and run:
+    ```bash
+    docker run --rm -v "$(pwd):/scan_target" nightly-cobweb-sweeper
+    ```
+    The `$(pwd)` part mounts your current working directory into the container at `/scan_target`. The `--rm` flag ensures the container is removed after the scan.
+
+    You can also specify a different target directory within the container if you mount it differently:
+    ```bash
+    docker run --rm -v /path/to/your/project:/my_project_dir nightly-cobweb-sweeper /my_project_dir
+    ```
+
+## Configuration (via Environment Variables)
+
+You can customize the scanner's behavior by passing environment variables to the `docker run` command:
+
+*   `MAX_FILE_SIZE_MB`: Maximum allowed file size in megabytes. Files larger than this will be flagged. Default: `10`.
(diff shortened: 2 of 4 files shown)


polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **Self‑contained container** – the Dockerfile builds a minimal Alpine image with only the tools the scanner needs (`bash`, `find`, `grep`, `coreutils`). This keeps the image small and reduces the attack surface.  
- **Clear entry‑point** – `ENTRYPOINT` points directly at `cobweb_scanner.sh` and `CMD` supplies a sensible default (`/scan_target`). Users can override the target directory without extra flags.  
- **Configurable via env vars** – `MAX_FILE_SIZE_MB` and `EXCLUDE_PATTERNS` are read from the environment, making the utility flexible for CI pipelines or ad‑hoc runs.  
- **Comprehensive detection** – the script covers temporary/backup files, empty files & directories, large files, and a handful of common sensitive filenames.  
- **Readable output** – the report is sectioned with clear headings, making it easy to pipe into other tools or grep for a specific category.  
- **Integration‑style test suite** – the test script builds the image, creates a deterministic filesystem, runs the container, and asserts on the exact output strings. This gives confidence that the end‑to‑end flow works as advertised.  

---

## 🧪 Tests
| Tests spin up a real Docker daemon and build the image each run. | Consider adding a **fast unit‑test layer** (e.g., using `bats-core` or plain `shunit2`) that sources `cobweb_scanner.sh` and calls its functions directly. This speeds up CI and isolates script logic from Docker. |
| Temporary directories are created with hard‑coded names (`temp_cobweb_test_dir`). | Use `mktemp -d` to generate a unique, automatically‑cleaned directory, e.g.:<br>`TEST_DIR=$(mktemp -d -t cobweb_test_XXXX)`<br>and remove it in the `cleanup` trap. |
| The script uses `set -euo pipefail`, which is good, but the Docker build command discards its output (`> /dev/null`). | Capture the build log on failure to aid debugging:<br>`docker build -t "$IMAGE_NAME" . || { echo "Docker build failed"; cat build.log; exit 1; }` |
| Assertions compare raw strings that include the absolute container path (`/scan_target/...`). | Abstract the path prefix into a variable (`SCAN_ROOT="/scan_target"`), then build expected strings with `${SCAN_ROOT}`. This makes the test resilient if the mount point changes. |
| No test for **error handling** (e.g., missing target directory, malformed `EXCLUDE_PATTERNS`). | Add a negative‑case test that runs the container without mounting a directory and asserts that the script exits with a non‑zero status and prints a helpful error. |
| Large‑file size check uses `bc` for floating‑point conversion. | Verify that `bc` is present in the Alpine image (it isn’t by default). Either add `bc` to the `apk add` line or replace the conversion with Bash arithmetic (`size_mb=$(awk "BEGIN {printf \"%.2f\", $SIZE_KB/1024}" )`). |
| The test suite builds the image with the tag `nightly-cobweb-sweeper-test`. | Ensure the tag is unique per CI run (e.g., include `$GITHUB_RUN_ID`) to avoid collisions when tests run in parallel. |

---

## 🔒 Security
- **Run as non‑root** – The container currently runs as the default `root` user. Add a low‑privilege user in the Dockerfile to limit the impact of a compromised script:<br>`RUN addgroup -S scanner && adduser -S scanner -G scanner`<br>`USER scanner` |
- **Limit mount permissions** – The README suggests mounting any directory. Encourage read‑only mounts when only scanning is needed: `-v "$(pwd):/scan_target:ro"` – this prevents accidental writes from the container.  
- **Sanitize `EXCLUDE_PATTERNS`** – The pattern matching uses Bash globbing (`[[ "$file" == $pattern ]]`). If a malicious pattern contains characters like `*` or `?` it could unintentionally match more than intended, but it cannot execute code. Still, document that patterns are **glob‑style**, not regex, and advise quoting them.  
- **Avoid leaking host paths** – The script prints absolute paths inside the container (`/scan_target/...`). This is fine, but be aware that if the container is run with privileged mounts, the output could reveal host directory structures. Consider an option to strip the prefix before printing.  
- **Dependency hygiene** – Alpine’s `apk add` pulls the latest versions at build time. Pin versions (e.g., `apk add --no-cache bash=5.2.15-r0`) to guarantee reproducible builds and avoid supply‑chain surprises.  

---

- **README enhancements**  
  - Add a **sample output** block so users can see the exact format.  
  - Document the **default values** (`MAX_FILE_SIZE_MB=10`, `EXCLUDE_PATTERNS` empty) in a table for quick reference.  
  - Explain the **non‑root recommendation** and show the `:ro` mount flag.  
  - Provide a one‑liner for CI usage, e.g., `docker run --rm -e MAX_FILE_SIZE_MB=5 -v "$PWD:/scan_target:ro" nightly-cobweb-sweeper`.  
- **Versioning** – Include a `VERSION` label in the Dockerfile (`LABEL version="0.1.0"`). This helps downstream users pin a specific release.  
- **Contribution guide** – Since the utility lives under `docker-tools/`, a short note on how to add new detection categories (e.g., “Add a new `find` command, pipe through `should_exclude`, and update the README”) would lower the barrier for future contributors.  
- **Shellcheck compliance** – Run `shellcheck` on `cobweb_scanner.sh` and fix any warnings (e.g., quoting variables, using `local` inside functions). This improves maintainability.  

---

## 🧱 Mocks / Fakes
- The current test suite **does not use any mocks**; it performs a full Docker build/run. This is acceptable for an integration test, but it makes the test suite slower and more fragile on machines without Docker.  
- **Potential mock strategies**  
  - **Mock `docker` commands** with a lightweight wrapper (e.g., a Bash function that records arguments) when running unit tests. This allows you to verify that the correct image name and volume mounts are used without actually pulling the daemon.  
  - Use **`testcontainers`** (if the repo already uses a language with such a library) to spin up a temporary Docker environment in CI, providing better isolation and automatic cleanup.  
  - For the script itself, you could **source a mock `find`** that returns a deterministic list of paths, enabling pure‑shell unit tests of the reporting logic.  

Implementing a small unit‑test layer with mocks will keep the fast feedback loop tight while still retaining the full integration test for end‑to‑end validation.

PR #5754, opened 2026-08-24
feat(gemini): A whimsical Node.js utility to send and receive encrypted messages across simulated cosmic distances, with configurable signal degradation.

polsala:
## Implementation Summary
- **Utility**: nightly-cosmic-comm-relay
- **Provider**: gemini
- **Location**: `node-utils/nightly-nightly-cosmic-comm-relay-5`
- **Files Created**: 3
- **Description**: A whimsical Node.js utility to send and receive encrypted messages across simulated cosmic distances, with configurable signal degradation.

## Rationale

## Why safe to merge
- Utility is isolated to `node-utils/nightly-nightly-cosmic-comm-relay-5`.

## Test Plan
- Follow the instructions in the generated README at `node-utils/nightly-nightly-cosmic-comm-relay-5/README.md`
- Run tests located in `node-utils/nightly-nightly-cosmic-comm-relay-5/tests/`

## Links

## Mock Justification

### DIFF
--- a/node-utils/nightly-nightly-cosmic-comm-relay-5/README.md
+++ b/node-utils/nightly-nightly-cosmic-comm-relay-5/README.md
@@ -1,0 +1,83 @@
+# Nightly Cosmic Comm Relay
+
+This utility simulates a whimsical cosmic communication relay. It allows you to send encrypted messages across vast, simulated distances, with the added fun of signal degradation that can affect message integrity. It's built with Node.js for cross-platform compatibility and ease of use.
+
+## Philosophy
+
+"Anarchy with discipline" — this tool is designed to be fun, slightly chaotic, and yet reliably functional. It embraces the spirit of exploration and communication, even in the face of cosmic interference.
+
+## Features
+
+*   **End-to-end Encryption**: Uses a simple XOR cipher for message security.
+*   **Simulated Cosmic Distance**: Messages take time to travel, adjustable via a `delay` parameter.
+*   **Signal Degradation**: Introduces random bit flips to simulate interference, controllable via a `degradationRate`.
+*   **Cross-Platform**: Runs on any system with Node.js installed.
+
+## Installation
+
+1.  Clone the repository:
+    ```bash
+    git clone https://github.com/polsala/ApocalypsAI.git
+    cd ApocalypsAI
+    ```
+2.  Navigate to the utility directory:
+    ```bash
+    cd utils/nightly-cosmic-comm-relay
+    ```
+3.  Install dependencies:
+    ```bash
+    npm install
+    ```
+
+## Usage
+
+### Sending a Message
+
+To send a message, run the `send` command:
+
+```bash
+node src/relay.js send --message "Hello from Earth!" --recipient "Proxima Centauri" --key "secretkey" --distance 5 --degradationRate 0.05
+```
+
+*   `--message`: The plaintext message to send.
+*   `--recipient`: The destination of the message (for display purposes).
+*   `--key`: The encryption key (a simple string).
+*   `--distance`: The simulated distance in light-years (affects travel time).
+*   `--degradationRate`: The probability of a bit flip per character (0.0 to 1.0).
+
+### Receiving a Message
+
+To simulate receiving a message, run the `receive` command. This will process any messages that have "arrived" based on their simulated travel time.
+
+```bash
+node src/relay.js receive --key "secretkey"
+```
+
+*   `--key`: The encryption key used to decrypt messages.
+
+### Example Workflow
+
+**Terminal 1 (Sender):**
+```bash
+node src/relay.js send --message "Greetings, fellow travelers!" --recipient "Andromeda Galaxy" --key "stardust" --distance 10 --degradationRate 0.1
+```
+
+**Terminal 2 (Receiver):**
+
+Wait for the simulated travel time (based on `--distance`). Then run:
+
+```bash
+node src/relay.js receive --key "stardust"
+```
+
+## Testing
+
+Run the tests using:
+
+```bash
+npm test
+```
+
+## Contributing
+
+Feel free to suggest improvements or new features! This is a whimsical project, so creativity is encouraged.
--- a/node-utils/nightly-nightly-cosmic-comm-relay-5/src/relay.js
+++ b/node-utils/nightly-nightly-cosmic-comm-relay-5/src/relay.js
@@ -1,0 +1,154 @@
+const crypto = require('crypto');
+const process = require('process');
+
+// --- Configuration ---
+const MAX_SIMULATED_DISTANCE = 100; // Light-years
+const MIN_SIMULATED_DISTANCE = 1;   // Light-years
+const DEFAULT_DEGRADATION_RATE = 0.01; // 1% chance of bit flip per character
+const DEFAULT_DELAY_MULTIPLIER = 500; // ms per light-year for travel time
+
+// --- Encryption/Decryption --- 
+function encrypt(text, key) {
+    const iv = crypto.randomBytes(16);
+    const cipher = crypto.createCipheriv('aes-256-cbc', Buffer.from(key, 'utf8'), iv);
+    let encrypted = cipher.update(text, 'utf8', 'hex');
+    encrypted += cipher.final('hex');
+    return `${iv.toString('hex')}:${encrypted}`;
+}
+
+function decrypt(text, key) {
+    const parts = text.split(':');
+    const iv = Buffer.from(parts.shift(), 'hex');
+    const encryptedText = parts.join(':');
+    const decipher = crypto.createDecipheriv('aes-256-cbc', Buffer.from(key, 'utf8'), iv);
+    let decrypted = decipher.update(encryptedText, 'hex', 'utf8');
+    decrypted += decipher.final('utf8');
+    return decrypted;
+}
+
+// --- Signal Degradation ---
+function degradeSignal(encryptedMessage, degradationRate) {
+    if (degradationRate <= 0) return encryptedMessage;
+
+    const messageBuffer = Buffer.from(encryptedMessage, 'hex');
+    const degradedBuffer = Buffer.from(messageBuffer);
+
+    for (let i = 0; i < degradedBuffer.length; i++) {
+        for (let bit = 0; bit < 8; bit++) {
+            if (Math.random() < degradationRate) {
+                degradedBuffer[i] ^= (1 << bit); // Flip a random bit
+            }
+        }
+    }
+    return degradedBuffer.toString('hex');
+}
+
+// --- Message Storage (simulated) ---
+// In a real scenario, this would be a database or persistent storage.
+// For this utility, we'll use a simple in-memory array, keyed by arrival time.
+const messageQueue = [];
+
+// --- Command Line Argument Parsing ---
+function parseArgs(args) {
+    const result = {};
+    for (let i = 0; i < args.length; i++) {
+        if (args[i].startsWith('--')) {
+            const key = args[i].substring(2);
+            const value = (i + 1 < args.length && !args[i+1].startsWith('--')) ? args[++i] : true;
+            result[key] = value;
+        }
+    }
+    return result;
+}
+
+// --- Core Logic ---
+async function sendMessage(options) {
+    const { message, recipient, key, distance, degradationRate } = options;
+
+    if (!message || !key) {
+        console.error("Error: Message and key are required for sending.");
+        process.exit(1);
+    }
+
+    const effectiveDistance = Math.max(MIN_SIMULATED_DISTANCE, Math.min(MAX_SIMULATED_DISTANCE, parseInt(distance || MIN_SIMULATED_DISTANCE)));
+    const effectiveDegradationRate = Math.max(0, Math.min(1, parseFloat(degradationRate || DEFAULT_DEGRADATION_RATE)));
+
+    const encryptedMessage = encrypt(message, key);
+    const degradedEncryptedMessage = degradeSignal(encryptedMessage, effectiveDegradationRate);
+
+    const travelTime = effectiveDistance * DEFAULT_DELAY_MULTIPLIER;
+    const arrivalTime = Date.now() + travelTime;
+
+    messageQueue.push({
+        id: Date.now() + Math.random(), // Simple unique ID
+        recipient: recipient || "Unknown Destination",
+        encryptedMessage: degradedEncryptedMessage,
+        arrivalTime: arrivalTime,
+        originalDistance: effectiveDistance,
+        degradationRate: effectiveDegradationRate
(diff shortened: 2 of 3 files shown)


polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **Clear intent & whimsical tone** – the README does a great job of setting the “cosmic” theme while still explaining the core features (encryption, delay, degradation).  
- **Modular helpers** – the code separates concerns nicely (`encrypt`, `decrypt`, `degradeSignal`, `parseArgs`, `sendMessage`, `receiveMessage`). This makes future extensions (e.g., different ciphers) straightforward.  
- **CLI ergonomics** – argument parsing is simple and works for both `--flag value` and boolean flags, which keeps the command line usage lightweight.  
- **Test scaffolding** – the test file already covers the main happy‑path scenarios (send, receive, error handling) and uses `exec` to run the script, which mirrors real‑world usage.

---

## 🧪 Tests
| **Process isolation** – each `runRelay` spawns a new Node process, so the in‑memory `messageQueue` is reset between the `send` and `receive` steps. The current tests therefore cannot actually retrieve a previously‑sent message. | Refactor the storage layer to be **process‑independent**. A lightweight JSON file (e.g., `./data/queue.json`) or a simple SQLite DB would persist messages across CLI invocations. Then the tests can assert on real receipt without needing to parse stdout from the `send` run. |
| **Key length handling** – `encrypt`/`decrypt` feed the raw UTF‑8 key directly into `crypto.createCipheriv`. AES‑256‑CBC expects a 32‑byte key; passing a shorter string throws `Invalid key length`. The tests use keys like `"testkey"` (8 bytes) which will cause runtime errors. | Derive a proper 32‑byte key from the user‑provided string, e.g.: <br>```js\nconst hash = crypto.createHash('sha256').update(key).digest();\nconst cipher = crypto.createCipheriv('aes-256-cbc', hash, iv);\n``` <br>Do the same in `decrypt`. Add a unit test that verifies encryption/decryption works with arbitrary‑length keys. |
| **Degradation API mismatch** – `degradeSignal` expects a hex string, but `encrypt` returns `iv:hexCipher`. The function is currently called on the whole `iv:payload` string, which will corrupt the IV and make decryption impossible. | Split the encrypted payload, degrade only the ciphertext part, then recombine: <br>```js\nconst [ivHex, cipherHex] = encrypted.split(':');\nconst degraded = degradeSignal(cipherHex, rate);\nreturn `${ivHex}:${degraded}`;\n``` <br>Add a test that sends a message with a non‑zero `degradationRate` and asserts that the receiver either successfully decrypts (low rate) or reports a degraded message (high rate). |
| **Error handling via `process.exit`** – the core functions call `process.exit(1)` on missing arguments. This terminates the whole process, making it hard to test error paths (the test suite would abort). | Throw an `Error` instead and let the CLI wrapper catch it, print a friendly message, and then call `process.exit`. Example: <br>```js\nif (!message || !key) throw new Error('Message and key are required');\n``` <br>In the CLI entry point: <br>```js\ntry { await sendMessage(opts); }\ncatch (e) { console.error('Error:', e.message); process.exit(1); }\n``` <br>Update the tests to expect the error message on `stderr` without the process terminating prematurely. |
| **Test reliability** – the tests mock `Date.now` but never reset the mock between individual `it` blocks (except in `afterAll`). This can cause cross‑test contamination. | Use `beforeEach`/`afterEach` to reset `mockDateNow` and restore `Date.now` for each test, ensuring deterministic timing. |
| **Missing coverage** – there is no test for the **edge case** where `degradationRate` is `0` or `1`, nor for invalid numeric inputs (negative distance, non‑numeric degradation). | Add parameter‑validation tests that feed malformed values and assert the CLI exits with a helpful error. |

---

## 🔒 Security
- **Key derivation** – feeding a raw UTF‑8 string into AES‑256‑CBC is insecure and will crash for short keys. Derive a fixed‑size key (e.g., SHA‑256 hash) as shown above.
- **IV reuse** – the current implementation generates a fresh IV per encryption, which is good. Ensure the IV is **always** transmitted alongside the ciphertext (the `iv:payload` format does this).
- **Degradation before decryption** – flipping bits in the ciphertext (or IV) will cause `crypto` to throw a `Decipher.final` error. The code should catch this and surface a graceful “message degraded” notice rather than bubbling an exception. Wrap decryption in a `try/catch` and return a sentinel value when it fails.
- **Dependency hygiene** – the only external dependency is Node’s built‑in `crypto`, which is fine. However, the project should lock the Node engine version in `package.json` (e.g., `"engines": {"node": ">=18"}`) to guarantee the required crypto primitives are available.
- **Input sanitisation** – the CLI accepts arbitrary strings for `recipient` and `message`. While they are not persisted beyond memory, if you later switch to a file‑based queue, ensure you escape or validate these fields to avoid injection attacks (e.g., when writing to a JSON file).

---

- **README path mismatch** – the installation steps point to `utils/nightly-cosmic-comm-relay`, but the actual folder is `node-utils/nightly-nightly-cosmic-comm-relay-5`. Update the instructions to reflect the real path or provide a top‑level alias script.
- **Usage examples vs. implementation** – the README shows commands like `node src/relay.js send …`, which matches the script, but the argument parser expects flags *after* the command (e.g., `node src/relay.js send --message "…"`). Clarify this ordering in the docs.
- **Help output** – consider adding a `--help` flag that prints a concise usage summary. This improves DX for new users and can be generated automatically from the `parseArgs` schema.
- **Error messages** – the CLI currently prints generic “Error: …” messages via `console.error`. Prefix them with a consistent tag (e.g., `Error:`) and exit with a non‑zero status. Document these error codes in the README.
- **Testing instructions** – the README says `npm test`, but the repo does not show a `test` script in `package.json`. Add a script entry: `"test": "jest"` (or `"test": "node ./node_modules/.bin/jest"`). Also list any required dev dependencies (`jest`, `cross-env` if needed).

---

## 🧱 Mocks / Fakes
- **Date mocking** – the test suite replaces `Date.now` globally, which is fine for a single test file but can interfere with other modules that rely on the real clock (e.g., `crypto.randomBytes`). Scope the mock to the module under test using `jest.spyOn(Date, 'now')` and restore it in `afterEach`.
- **Message queue persistence** – because the queue lives only in memory, the tests resort to “pragmatic” stdout parsing. A better approach is to **inject** the storage layer. Export a `createRelay({ storage })` factory where `storage` implements `add(message)` and `popReady(now)`. In production you pass the in‑memory array; in tests you pass a mock object that records calls, enabling precise assertions without spawning separate processes.
- **Crypto randomness** – the high degradation test relies on the real `crypto.randomBytes` for IV generation, which makes the output nondeterministic. For deterministic unit tests, mock `crypto.randomBytes` to return a fixed buffer (e.g., `Buffer.alloc(16, 0)`). This will make the encrypted payload stable and allow you to assert on the exact degraded output.

---

### Quick win checklist
- [ ] Derive a 32‑byte key from the user‑provided string (SHA‑256) and update encryption/decryption accordingly.  
- [ ] Persist the message queue to a file (JSON) so that separate CLI invocations can share state.  
- [ ] Refactor error handling to throw `Error` objects instead of calling `process.exit` inside core functions.  
- [ ] Add validation for numeric arguments (`distance`, `degradationRate`) and unit tests for invalid inputs.  
- [ ] Update the README paths, add a `--help` flag, and document the new `npm test` script.  
- [ ] Adjust the degradation logic to only corrupt the ciphertext portion, preserving the IV.  
- [ ] Introduce a storage abstraction to make the queue mockable in tests, eliminating the need for cross‑process stdout parsing.  

Implementing these changes will make the utility more robust, secure, and testable while preserving its fun, cosmic charm. 🚀✨

polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **Clear purpose & fun narrative** – the utility’s “cosmic” theme is well‑communicated in the README and makes the project approachable.  
- **Modular helpers** – encryption (`encrypt`/`decrypt`), degradation (`degradeSignal`), argument parsing (`parseArgs`) and the in‑memory queue are each isolated in their own functions, which eases future refactoring.  
- **CLI entry point** – the script can be invoked with `node src/relay.js send …` and `node src/relay.js receive …`, matching the usage examples.  
- **Test scaffolding** – a fairly complete Jest suite is provided, covering happy‑path sends, custom distance/degradation, receipt, and error handling.  

## 🧪 Tests
| Issue | Why it matters | Suggested fix |
|-------|----------------|---------------|
| **Process isolation** – each `runRelay` spawns a new Node process, so the in‑memory `messageQueue` and the mocked `Date.now` in the test process are not visible to the child. | Tests that send a message and then call `receive` will always see an empty queue, causing false failures. | Persist the queue to a temporary file (e.g. JSON in `os.tmpdir()`) or expose a simple IPC (e.g. `process.send`) so the child can read the same state. <br>```js// src/relay.js  // replace in‑memory queue\nconst fs = require('fs');\nconst path = require('path');\nconst QUEUE_FILE = path.join(require('os').tmpdir(), 'cosmic-queue.json');\nfunction loadQueue() { try { return JSON.parse(fs.readFileSync(QUEUE_FILE)); } catch (_) { return []; } }\nfunction saveQueue(q) { fs.writeFileSync(QUEUE_FILE, JSON.stringify(q)); }\n``` |
| **Mocking `Date.now` ineffective** – the mock only affects the test runner, not the spawned script. | Arrival‑time calculations become nondeterministic, breaking the “message arrived” assertions. | Pass the current timestamp via an environment variable (e.g. `MOCK_NOW=123456`) and have the script read `process.env.MOCK_NOW` when present. <br>```js// src/relay.js\nconst now = process.env.MOCK_NOW ? Number(process.env.MOCK_NOW) : Date.now();``` |
| **Random degradation** – `degradeSignal` flips bits based on `Math.random()`. | High‑degradation tests can produce different ciphertexts on each run, making the expected “DEGRADED/UNREADABLE” output flaky. | Seed the RNG in tests (e.g. use `seedrandom`) or replace `Math.random` with a deterministic stub when `process.env.TEST_MODE` is set. |
| **CLI argument parsing** – `parseArgs` ignores the first positional argument (`send`/`receive`). | Options are parsed correctly, but the command itself is lost, so the script may treat `send` as an unknown flag. | Extract the command before calling `parseArgs`.<br>```js\nconst [cmd, ...rest] = process.argv.slice(2);\nconst opts = parseArgs(rest);\nif (cmd === 'send') await sendMessage(opts);\nelse if (cmd === 'receive') await receiveMessages(opts);\nelse console.error('Unknown command');\n``` |
| **Missing assertions on exit codes** – the script calls `process.exit(1)` on validation errors, but the tests only inspect `stderr/stdout`. | A failing test could silently pass if the process exits with a non‑zero code that `exec` treats as an error. | Capture `error?.code` in the test helper and assert it matches expectations. |
| **Test reliability** – several expectations rely on exact phrasing (`Message sent to …!`). If the script’s wording changes, many tests break. | Tight coupling to output strings makes refactoring harder. | Export a small API (e.g. `module.exports = { sendMessage, receiveMessages }`) and test the functions directly, reserving a few integration tests for the CLI. |

## 🔒 Security
- **Key handling** – `crypto.createCipheriv('aes-256-cbc', Buffer.from(key, 'utf8'), iv)` expects a 32‑byte key. Supplying a short string (e.g. `"testkey"` in the tests) throws `Invalid key length`.  
  - **Fix**: Derive a 256‑bit key from the user‑provided passphrase, e.g. using SHA‑256 or PBKDF2.<br>```js\nfunction deriveKey(pass) {\n  return crypto.createHash('sha256').update(pass).digest();\n}\nconst keyBuf = deriveKey(key);\nconst cipher = crypto.createCipheriv('aes-256-cbc', keyBuf, iv);\n```  
- **No integrity/authentication** – AES‑CBC provides confidentiality only; a corrupted ciphertext (from degradation) will raise a padding error or produce garbled plaintext. Consider switching to an authenticated mode (AES‑GCM) or appending an HMAC.  
- **Signal degradation on ciphertext** – flipping bits arbitrarily destroys the AES padding and can cause decryption to throw. The current `receive` flow should catch such errors and present a friendly “DEGRADED/UNREADABLE” message, but the catch block is missing.  
  - **Fix**: Wrap `decrypt` in a `try/catch` and return a sentinel value on failure.  
- **IV reuse** – The IV is generated per message, which is good, but the key derivation step must guarantee a constant length; otherwise the same IV could be reused with a different‑length key, weakening security.  
- **Dependency hygiene** – The repo does not list `crypto` (built‑in) but does not declare `jest` or any test runner in `package.json`. Ensure `devDependencies` include `"jest": "^29.0.0"` and a `"test"` script (`"npm test": "jest"`).  

- **Path mismatch** – README installation steps reference `cd utils/nightly-cosmic-comm-relay`, while the actual directory is `node-utils/nightly-nightly-cosmic-comm-relay-5`. Update the docs to reflect the real path or provide a symlink.  
- **Missing npm scripts** – The README mentions `npm test` but the repository lacks a `"test"` script. Add to `package.json`:<br>```json\n\"scripts\": { \"test\": \"jest\" }\n```  
- **CLI help** – Users have no way to discover available flags. Implement a `--help` flag that prints usage (e.g. using `commander` or a simple manual block).  
- **Environment variables** – Document the optional `MOCK_NOW` and `TEST_MODE` variables used for deterministic testing.  
- **Example output** – The README shows a multi‑line message format; ensure the actual script prints exactly that (including line breaks) or adjust the docs.  
- **License & contribution guidelines** – Add a `LICENSE` file and a brief `CONTRIBUTING.md` to encourage community involvement.  

## 🧱 Mocks / Fakes
- **Date mocking** – As noted, mocking `Date.now` in the test process does not affect the child process. Use an explicit injection (env var) or refactor the timing logic into a separate module that can be required and mocked with Jest’s `jest.mock`.  
- **Message queue persistence** – For deterministic tests, replace the in‑memory `messageQueue` with a lightweight file‑backed store (JSON) that both the CLI and the test harness can read/write. This also makes the utility more realistic (messages survive process restarts).  
- **Crypto stubbing** – High‑degradation tests rely on decryption failure. Instead of relying on random bit flips, stub `crypto.createDecipheriv` to throw an error when a flag (e.g. `process.env.FORCE_DEGRADE`) is set. This yields deterministic “unreadable” outcomes.  
- **CLI execution wrapper** – The current `runRelay` helper swallows `error` from `exec`. Return the `error` object (including `code` and `signal`) so tests can assert on non‑zero exit codes.  

---

### Quick win checklist
1. **Derive a 32‑byte key** from the user‑supplied passphrase.  
2. **Persist the queue** to a temporary JSON file and read/write it in both send/receive.  
3. **Expose the command** (`send`/`receive`) before parsing flags.  
4. **Add a deterministic time source** (`MOCK_NOW` env var) and update tests accordingly.  
5. **Wrap decryption in `try/catch`** and surface a friendly degraded‑message notice.  
6. **Update README paths**, add npm test script, and document the new env vars.  

Implementing these changes will make the utility reliable, secure, and pleasant to use and maintain. Happy coding!

PR #5785, opened 2026-08-27
feat(gemini): A whimsical bash utility that periodically checks and optionally synchronizes system time with an NTP server, logging any significant temporal deviations as 'anomalies'.

polsala:
## Implementation Summary
- **Utility**: nightly-chrono-sync-beacon
- **Provider**: gemini
- **Location**: `bash-utils/nightly-nightly-chrono-sync-beacon`
- **Files Created**: 3
- **Description**: A whimsical bash utility that periodically checks and optionally synchronizes system time with an NTP server, logging any significant temporal deviations as 'anomalies'.

## Rationale

## Why safe to merge
- Utility is isolated to `bash-utils/nightly-nightly-chrono-sync-beacon`.

## Test Plan
- Follow the instructions in the generated README at `bash-utils/nightly-nightly-chrono-sync-beacon/README.md`
- Run tests located in `bash-utils/nightly-nightly-chrono-sync-beacon/tests/`

## Links

## Mock Justification

### DIFF
--- a/bash-utils/nightly-nightly-chrono-sync-beacon/README.md
+++ b/bash-utils/nightly-nightly-chrono-sync-beacon/README.md
@@ -1,0 +1,61 @@
+# Nightly Chrono-Sync Beacon
+
+## Summary
+The `nightly-chrono-sync-beacon` is a whimsical-yet-useful bash utility designed to maintain temporal stability on your system. It periodically queries a designated NTP (Network Time Protocol) server to detect any 'temporal anomalies' (significant time drift) and can optionally synchronize your system's clock to ensure cosmic alignment.
+
+## Features
+- **Temporal Drift Detection**: Compares local system time with a reliable NTP source.
+- **Anomaly Logging**: Logs any detected time deviations exceeding a configurable threshold.
+- **Optional Synchronization**: Can automatically correct system time if a significant drift is found (requires `sudo` privileges for `sntp`).
+- **Configurable**: Easy to adjust NTP server and drift threshold.
+
+## Usage
+
+### Prerequisites
+- `bash` (most Linux/macOS systems have this)
+- `sntp` command (part of `ntp` or `ntpsec` packages, or `ntpdate` on older systems). On Debian/Ubuntu: `sudo apt install ntpdate` or `sudo apt install ntpsec-ntpclient`.
+- `bc` for floating-point comparisons (usually pre-installed, or `sudo apt install bc`).
+
+### Running the Beacon
+1. Make the script executable:
+   ```bash
+   chmod +x src/chrono_sync_beacon.sh
+   ```
+2. Run it directly:
+   ```bash
+   ./src/chrono_sync_beacon.sh
+   ```
+
+### Configuration
+You can configure the beacon using environment variables:
+
+- `NTP_SERVER`: The NTP server to query. Defaults to `pool.ntp.org`.
+  Example: `NTP_SERVER="time.google.com" ./src/chrono_sync_beacon.sh`
+
+- `DRIFT_THRESHOLD_SECONDS`: The maximum acceptable time difference in seconds before a 'temporal anomaly' is logged. Defaults to `5`.
+  Example: `DRIFT_THRESHOLD_SECONDS="1.5" ./src/chrono_sync_beacon.sh`
+
+- `SYNC_ENABLED`: Set to `true` to enable automatic time synchronization if a significant drift is detected. Defaults to `false`. **Note: Synchronization requires `sudo` privileges for the `sntp` command.**
+  Example: `SYNC_ENABLED="true" sudo ./src/chrono_sync_beacon.sh`
+
+- `LOG_FILE`: Path to the log file. Defaults to `/var/log/chrono_sync_beacon.log`. Output is also sent to `stdout`.
+  Example: `LOG_FILE="/tmp/my_beacon.log" ./src/chrono_sync_beacon.sh`
+
+### Example Cron Job (for daily checks)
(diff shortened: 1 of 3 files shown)


polsala:
## 🤖 Review by GROQ Agent

## ✅ What's solid
- **Self‑contained utility** – All code lives under `bash-utils/nightly-nightly-chrono-sync-beacon`, so it won’t interfere with the rest of the repo.  
- **Clear defaults & configurability** – The script reads its configuration from environment variables with sensible fall‑backs (`pool.ntp.org`, 5 s threshold, logging to `/var/log/...`).  
- **Graceful pre‑flight checks** – The script validates the presence of `sntp` and `bc` before proceeding, emitting helpful error messages.  
- **Human‑readable logging** – Timestamped log lines are written both to stdout and to the configurable log file via `tee`.  
- **Comprehensive test harness** – The test suite covers the main decision matrix (no drift, small drift, large drift with/without sync, missing commands, parsing failures, etc.) and runs completely offline thanks to the mocks.  

## 🧪 Tests
| Area | Feedback |
|------|----------|
| **Coverage** | The tests hit the major branches, but a few edge‑cases are still untested: <br>• What happens when `bc` is missing? <br>• Behaviour when the log file is not writable (e.g., `/var/log/...` without root). |
| **Isolation** | Overriding the built‑in `command` function is clever, but it also masks *all* `command -v` checks (including the one for `bc`). Consider narrowing the mock to only the `sntp` lookup, e.g.: <br>```bash\ncommand() { if [[ $1 == -v && $2 == sntp ]]; then return ${_mock_sntp_exit_code}; else builtin command "$@"; fi }\n``` |
| **Readability** | The `run_test` helper is a bit dense. Splitting the assertions (`assert_exit_code`, `assert_contains`, …) into a separate `assert.sh` library would make each test case easier to scan. |
| **Exit‑code expectations** | Test 5 expects a non‑zero exit when sync fails, which is good. However the script currently exits with `1` only on pre‑flight failures; the sync‑failure path also returns `1`. Make sure this is intentional and documented. |
| **Test naming** | The test titles are clear, but adding a short comment on the *why* (e.g., “ensure we don’t attempt sudo when SYNC_ENABLED is false”) would aid future maintainers. |

## 🔒 Security
- **Privilege escalation** – The script calls `sudo sntp -s`. If an attacker can influence `NTP_SERVER` (e.g., via an environment variable they control), they could force the system to sync to a malicious NTP source. Mitigation: validate that `NTP_SERVER` matches a whitelist or at least a simple regex (`^[a-zA-Z0-9.-]+$`).  
- **Log file path** – The default `/var/log/chrono_sync_beacon.log` requires root write access. If the script is run as a regular user with `SYNC_ENABLED=true`, the `sudo sntp` will succeed but the `tee` to the log file will fail, potentially leaking sensitive timing information to stdout. Consider falling back to a user‑writable location (e.g., `$HOME/.chrono_sync_beacon.log`) when the default is not writable.  
- **Command injection** – The script passes `$NTP_SERVER` directly to `sntp`. While `sntp` treats its argument as a hostname, a malicious value containing spaces or shell metacharacters could be interpreted by the surrounding `bash` if the script ever expands it in a different context. Always quote the variable (`sntp -d "$NTP_SERVER"` – already done) and consider additional sanitisation.  
- **Mocking `command`** – In the test suite, redefining `command` globally could hide missing binaries in other scripts that might be sourced later. Scope the mock to the script under test (e.g., by using a subshell) to avoid accidental side‑effects.  

- **README enhancements**  
  - Add a **“Exit codes”** table so users know what `0`, `1`, etc., mean.  
  - Clarify the **required permissions** for each mode (read‑only vs. sync).  
  - Provide a quick‑start example that sets `LOG_FILE` to a user‑writable path for first‑time testing.  
- **Usage ergonomics**  
  - Consider adding a `--help` flag that prints the configurable env vars and defaults; this is more discoverable than scanning the README.  
  - Expose a `--once` flag (default) and a `--loop INTERVAL` flag for continuous monitoring without needing a cron job.  
- **Code style**  
  - Add `set -euo pipefail` at the top of the script to abort on unset variables and pipeline failures.  
  - Mark configuration variables as `readonly` after they are initialised to prevent accidental mutation later in the script.  
  - Use `printf` instead of `echo` for log lines to avoid issues with `-e` or backslashes.  

## 🧱 Mocks / Fakes
- **`sntp` mock** – The mock correctly simulates both the normal query (`-d`) and the sync (`-s`) paths, and it records whether a sync was attempted. This is solid.  
- **`sudo` mock** – Forwarding to the mocked `sntp` works, but it assumes *all* sudo calls are for `sntp`. If future changes add other privileged commands, the mock will silently pass them through to the real `sudo`. A safer approach is to explicitly check the arguments:  
  ```bash
  sudo() {
      if [[ "$1" == "sntp" ]]; then
          shift
          sntp "$@"
      else
          echo "Unexpected sudo call: $*" >&2
          return 1
      fi
  }
  ```  
- **`command` mock** – As noted above, the current implementation intercepts *any* `command -v` call. Tightening the mock to only the `sntp` lookup will reduce the risk of false positives in other parts of the test suite.  
- **Test isolation** – Each test resets the mock globals (`mock_sntp_output`, `mock_sntp_exit_code`, etc.) which is good. Consider wrapping the entire test run in a subshell (`( … )`) so that any accidental export or function leakage cannot affect the host shell.  

---

Overall, the utility is a fun and useful addition, and the test suite demonstrates a solid approach to offline testing of system‑level scripts. Addressing the minor portability, security, and documentation points above will make the tool more robust and easier for contributors to adopt. Happy syncing!

